PR #1, opened 2019-01-10
Initial app setup.

douglashall:
You can see the app running here: https://user-account.edxdev.org/


.travis.yml
```diff
@@ -0,0 +1,19 @@
+language: node_js
+node_js:
+  - lts/*
+cache:
+  directories:
+    - "~/.npm"
+before_install:
+  - npm install -g npm@latest
+  - npm install -g greenkeeper-lockfile@1.14.0
```
robrap: Are we officially using greenkeeper and is there anything additional we need to do to configure?

config/webpack.prod.config.js
```diff
@@ -0,0 +1,149 @@
...
+        use: [
+          MiniCssExtractPlugin.loader,
+          {
+            loader: 'css-loader', // translates CSS into CommonJS
+            options: {
+              sourceMap: true,
+              minimize: true,
+            },
+          },
+          'postcss-loader',
```
robrap: @staubina: FYI: This came from grades and doesn't exist in the cookie-cutter, but the footer won't work without this and there may be some browser issues.  You may not care for publisher, but it should be moved into the cookie-cutter.
@reillz10: Did you have plans to update the cookie-cutter or should we add a task?

.dockerignore
```diff
@@ -0,0 +1,10 @@
+node_modules
+npm-debug.log
+README.md
```
robrap: Looks like you removed the README file altogether?  We should have one and translate it to rst (automated with pandoc).  Let me know if you want me to take care of that.

config/webpack.dev.config.js
```diff
@@ -0,0 +1,139 @@
...
+  // Specify additional processing or side-effects done on the Webpack output bundles as a whole.
+  plugins: [
+    // Generates an HTML file in the output directory.
+    new HtmlWebpackPlugin({
+      inject: true, // Appends script tags linking to the webpack bundles at the end of the body
+      template: path.resolve(__dirname, '../public/index.html'),
+    }),
+    new webpack.EnvironmentPlugin({
+      NODE_ENV: 'development',
+      BASE_URL: 'localhost:1994',
```
robrap: We should update port 1994 to a unique port for eventual docker devstack integration.
Note: This port appears in multiple places, including the README when that is returned.

config/webpack.prod.config.js
staubina: I will try to keep this in mind.  We will eventually be adding the Header and Footer.

PR #3, opened 2019-01-16
Add Segment for analytics.

robrap:


src/segment.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // Add a version to keep track of what's in the wild.
+  analytics.SNIPPET_VERSION = '4.1.0';
+
+  // Load Analytics.js with your key, which will automatically
+  // load the tools you've enabled for your account. Boosh!
+  analytics.load(configuration.SEGMENT_KEY);
+
+  // TODO: Move this page call if we start making page calls
+  // - during routing.
```
robrap: @douglashall: FYI: I added this call to analytics.page() until we decide to call page from the app (using redux-beacon/segment or otherwise), so this PR stands on its own.

PR #6, opened 2019-01-18
Adding wireframe and layout

abutterworth:


src/components/UserAccount/index.jsx
```diff
@@ -115,9 +129,50 @@ class UserAccount extends React.Component {
...
             </Col>
             <Col md={8} lg={9} className="pt-md-1">
-
-              <h3>About {firstName}</h3>
-              <p>{aboutMe}</p>
+              <EditableContent
+                fieldName="about"
+                isEditing={this.state.currentlyEditingField === 'about'}
+                disabled={false}
+                renderStatic={props => ( // eslint-disable-line no-unused-vars
```
abutterworth: @douglashall Is this a crazy way to do editable fields? The EditableContent component takes 3 render props: renderStatic, renderEditable, renderEditing.

src/constants/education.js
```diff
@@ -0,0 +1,13 @@
+const EDUCATION = {
+  p: 'Doctorate',
```
albemarle: These should probably map to message ids like "common.education.phd" etc.  I can fix it up after merging, though.

src/constants/social.js
```diff
@@ -0,0 +1,34 @@
+const SOCIAL = {
+  linkedin: {
+    title: 'Linked In',
```
albemarle: Same here, "common.social.linkedin" etc.

src/actions/profileActions.js
```diff
@@ -0,0 +1,30 @@
+import AsyncActionType from './AsyncActionType';
```
douglashall: The naming convention we have been following for actions is:
```
src/actions/profile.js
```

src/services/UserProfileApiService.js
```diff
@@ -0,0 +1,65 @@
+import camelcaseKeys from 'camelcase-keys';
```
douglashall: Ideally, we would keep these functions together with the read functions to create a comprehensive API service layer for the user account API. The best place for that service layer is probably in a separate npm package and not `frontend-auth`. In order of preference, I think we either want to do that refactoring now, leave the write functions in frontend-auth for the time being, or leave the write functions here and remove them from frontend-auth.

src/data/reducers/ProfileReducer.js
```diff
@@ -0,0 +1,39 @@
+import { SAVE_USER_PROFILE } from '../../actions/profileActions';
```
douglashall: Can you remove the related reducer from `frontend-auth`?

src/actions/AsyncActionType.js
```diff
@@ -0,0 +1,34 @@
+/**
```
douglashall: Can you remove the related actions from frontend-auth?

src/sagas/RootSaga.js
```diff
@@ -0,0 +1,39 @@
...
+  saveUserProfileReset,
+} from '../actions/profileActions';
+
+import * as UserProfileApiService from '../services/UserProfileApiService';
+
+function* saveUserProfile(action) {
+  try {
+    yield put(saveUserProfileBegin());
+    const userAccount = yield call(
+      UserProfileApiService.saveUserProfile,
```
douglashall: Do we need to update the `userAccount` redux state with the updated data returned from the PATCH request to the user account API?

src/services/UserProfileApiService.js
davidjoy: I'm assuming that in the long run there's going to be a difference between APIs that are used to get user session data and APIs that are used to manage profile data for an arbitrary user.  In that sense - and because some of the refactoring called for pulling them out - it felt like the methods to save a user profile don't belong in a frontend-auth package.  

That said, it's also not clear to me that having all API methods available in a comprehensive package is going to be the right thing to do either.  Given those three possibilities, it felt most correct for them to live in the profile app, so I put them here as part of pulling the reducer and other profile-page specific code out of frontend-auth. 

We could refactor that further, but my understanding is that for v0.x we're not concerning ourselves with broader refactorings like that.

src/actions/profileActions.js
davidjoy: Happy to change it in the interest of current consistency.  That said, having unique filenames really helps when using "Go to file"-style commands in an IDE... would love to bring this up for discussion later when we're solidifying our patterns.

src/services/UserProfileApiService.js
douglashall: I think in the end it will make sense for fetch functions to live with save functions at the API service layer. If we don't want to do option #1, can we at least go with option #2 (leave the save functions in frontend-auth)?

src/sagas/RootSaga.js
davidjoy: Yes - that's done a few lines down from here where it manually creates a fetchUserAccountSuccess action.

davidjoy: Ideally the data would live in some common place where interested UIs could get to it, but for the foreseeable future - since the fetch user account and save user profile methods both return identical structures, I just let userAccount be the authority on that.

src/actions/profileActions.js
douglashall: Cool. Does your IDE's autocomplete not take into account directories?

src/sagas/RootSaga.js
douglashall: Ah, got it. Thanks for pointing it out.

src/data/reducers/ProfileReducer.js
davidjoy: Sure can.

src/actions/AsyncActionType.js
davidjoy: 👍

src/actions/profileActions.js
davidjoy: Does, I believe, but it's less typing without the "index" :)

src/services/UserProfileApiService.js
davidjoy: I guess I'm not quite clear why profile saving functions would go in an "auth" repository... is frontend-auth actually more of a frontend-api package?

davidjoy: I mean, I'm all for having them all in one place!  But it just seems like a misnomer, kinda.

douglashall: I totally agree that `frontend-auth` is not the right place for the `UserAccountApiService`, hence my #1 option above (refactor it to a separate `frontend-service-user-account` or whatever name npm package). I think it's better to keep everything together until we want to to that refactor and everything is already together in `frontend-auth`. For added context, the fetch interface is currently in use by other apps that consume `frontend-auth`, namely https://github.com/edx/edx-portal.

douglashall:
@davidjoy @abutterworth I rebased/squashed this. I think we are ready for merge. Sound good?

PR #16, opened 2019-02-14
Tests for the saga, AsyncActionType, and profile actions.

davidjoy:
Also changing the way the saga gets ahold of the UserAccountApiService to make it more testable.

### DIFF
--- a/src/actions/profile.test.js
+++ b/src/actions/profile.test.js
@@ -1,4 +1,8 @@
 import {
+  openEditableField,
+  closeEditableField,
+  EDITABLE_FIELD_OPEN,
+  EDITABLE_FIELD_CLOSE,
   SAVE_USER_PROFILE,
   saveUserProfileBegin,
   saveUserProfileSuccess,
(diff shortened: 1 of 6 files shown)


src/actions/profile.test.js
```diff
@@ -7,6 +11,24 @@ import {
...
+  });
+
+  it('should create a closed action', () => {
+    const expectedAction = {
+      type: EDITABLE_FIELD_CLOSE,
+      fieldName: 'name',
+    };
+    expect(closeEditableField('name')).toEqual(expectedAction);
+  });
+});
```
abutterworth: Are these types of tests useful? Seems like a lot of extra work/code over time when many action creators will look like this.

```
export const openEditableField = fieldName => ({
  type: EDITABLE_FIELD_OPEN,
  fieldName,
});
```

src/sagas/RootSaga.test.js
```diff
@@ -0,0 +1,121 @@
...
+        education: 'BS',
+        socialLinks: {
+          twitter: {
+            url: 'https://www.twitter.com',
+          },
+          facebook: {
+            url: 'https://www.facebook.com',
+          },
+        },
+      };
```
abutterworth: Try rebasing from master now that social links saving is fixed. The PROP_TO_STATE_MAP in RootSaga.js is no long expecting an object of socialLinks, but an array – mirroring the API.

src/actions/profile.test.js
davidjoy: Yup.  It tells us that fieldName was properly added to the object, and that we have the right constant in there.

When you add a dozen actions all at once with similar names and/or a dozen arguments, it's nice to double-check they're correct. :)

davidjoy: Furthermore, code without tests is much less likely to get tested when modified than code with existing tests.  It may seem a bit silly now, but we're helping people do the right thing later, too.

src/sagas/RootSaga.test.js
davidjoy: 👍 Just figured this out. :) Thanks!

src/actions/profile.test.js
abutterworth: 👍

PR #17, opened 2019-02-14
Is getting codecov to work as easy as this?

davidjoy:


albemarle:
I seem to also have gotten a coverage email... is there a way to turn those off?

PR #18, opened 2019-02-14
Add empty state cases for each section. 

abutterworth:
Added empty states for each section. Updated SwitchContent to handle forwarding of cases with a recursive function. This means you could create an infinite loop if you did something like:
```
<SwitchContent 
   expression="edit"
   cases={{
      edit: 'default',
      default: 'edit',
   }}
/>

### CODE REVIEW

File: src/components/UserProfile/SocialLinks.jsx [hunk 1]
```diff
@@ -83,7 +82,7 @@ class SocialLinks extends React.Component {
                 content="Social Links"
                 showEditButton
                 onClickEdit={() => onEdit('socialLinks')}
-                showVisibility={Boolean(socialLinks.length)}
+                showVisibility={Boolean(socialLinks && socialLinks.length)}
```
davidjoy: Stylistically I'd probably just do `socialLinks && socialLinks.length > 0` here and avoid the `Boolean`...

abutterworth: Good point that was leftover from elsewhere.

PR #22, opened 2019-02-20
Viewing others profiles and naming refactor.

davidjoy:
This PR allows users to view other user's profiles.  It also changes the path of our app to `/u/<username>`.  If you go to just `/` you'll get a simple 404 page.

This PR also has a refactoring of our actions/sagas/services.  Rather than continue to put profile-related API actions into the frontend-auth project, we decided (against prior decision) to put them in this app, as that feels more correct long term than overloading the API service in frontend-auth with things that aren't specific to auth.  This also let us unroll the service creation and decouple the services from the components (compare the way fetchUserAccount is called in the index.jsx file with the way the ProfileApiService can be called in the sagas).

Note that the “fetchProfile” API is hitting the same ‘accounts’ API as `UserAccountApiService.fetchUserAccount` from frontend-auth.  This is an intentional abstraction for clarity of purpose and assuming they may diverge. It also lets us group all the profile app’s API methods together, letting the ‘session’ API stay in frontend-auth.

We also decided to do some renaming:
- “user profile” -> “profile” cause there’s no other type of profile here.
- The saga handlers are prefixed with “handle”
- The API methods use proper HTTP verbs.
- The actions keep fetch and save as verbs as they’re clearer in components than “get” and “patch”
- The goal here is to differentiate different parts of the code, making it easier to remember where you are.

### DIFF
--- a/config/webpack.common.config.js
+++ b/config/webpack.common.config.js
@@ -9,6 +9,7 @@ module.exports = {
   },
   output: {
     path: path.resolve(__dirname, '../dist'),
+    publicPath: '/',
   },
   resolve: {
     extensions: ['.js', '.jsx'],
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -9,6 +9,7 @@ const commonConfig = require('./webpack.common.config.js');
 
 module.exports = Merge.smart(commonConfig, {
   mode: 'development',
+  devtool: 'eval-source-map',
   entry: [
     // enable react's custom hot dev client so we get errors reported in the browser
     require.resolve('react-dev-utils/webpackHotDevClient'),
@@ -24,6 +25,10 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
+          path.resolve(__dirname, 'node_modules/camelcase-keys'),
+          path.resolve(__dirname, 'node_modules/camelcase'),
+          path.resolve(__dirname, 'node_modules/map-obj'),
+          path.resolve(__dirname, 'node_modules/quick-lru'),
         ],
         loader: 'babel-loader',
         options: {
@@ -136,5 +141,6 @@ module.exports = Merge.smart(commonConfig, {
     historyApiFallback: true,
     hot: true,
     inline: true,
+    publicPath: '/',
   },
 });
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -24,6 +24,10 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
+          path.resolve(__dirname, 'node_modules/camelcase-keys'),
+          path.resolve(__dirname, 'node_modules/camelcase'),
+          path.resolve(__dirname, 'node_modules/map-obj'),
+          path.resolve(__dirname, 'node_modules/quick-lru'),
         ],
         loader: 'babel-loader',
       },
--- a/src/actions/profile.js
+++ b/src/actions/profile.js
@@ -2,9 +2,10 @@ import AsyncActionType from './AsyncActionType';
 
 export const EDITABLE_FIELD_OPEN = 'EDITABLE_FIELD_OPEN';
 export const EDITABLE_FIELD_CLOSE = 'EDITABLE_FIELD_CLOSE';
-export const SAVE_USER_PROFILE = new AsyncActionType('PROFILE', 'SAVE_USER_PROFILE');
-export const SAVE_USER_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'SAVE_USER_PROFILE_PHOTO');
-export const DELETE_USER_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'DELETE_USER_PROFILE_PHOTO');
+export const FETCH_PROFILE = new AsyncActionType('PROFILE', 'FETCH_PROFILE');
+export const SAVE_PROFILE = new AsyncActionType('PROFILE', 'SAVE_PROFILE');
+export const SAVE_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'SAVE_PROFILE_PHOTO');
+export const DELETE_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'DELETE_PROFILE_PHOTO');
 
 export const openEditableField = fieldName => ({
   type: EDITABLE_FIELD_OPEN,
@@ -16,25 +17,48 @@ export const closeEditableField = fieldName => ({
   fieldName,
 });
 
-export const saveUserProfileBegin = () => ({
-  type: SAVE_USER_PROFILE.BEGIN,
+export const fetchProfileBegin = () => ({
+  type: FETCH_PROFILE.BEGIN,
 });
 
-export const saveUserProfileSuccess = () => ({
-  type: SAVE_USER_PROFILE.SUCCESS,
+export const fetchProfileSuccess = profile => ({
+  type: FETCH_PROFILE.SUCCESS,
+  payload: { profile },
 });
 
-export const saveUserProfileReset = () => ({
-  type: SAVE_USER_PROFILE.RESET,
-});
-
-export const saveUserProfileFailure = error => ({
-  type: SAVE_USER_PROFILE.FAILURE,
+export const fetchProfileFailure = error => ({
+  type: FETCH_PROFILE.FAILURE,
   payload: { error },
 });
 
-export const saveUserProfile = (username, userAccountState, fieldName) => ({
-  type: SAVE_USER_PROFILE.BASE,
+export const fetchProfileReset = () => ({
+  type: FETCH_PROFILE.RESET,
+});
+
+export const fetchProfile = username => ({
+  type: FETCH_PROFILE.BASE,
+  payload: { username },
+});
+
+export const saveProfileBegin = () => ({
+  type: SAVE_PROFILE.BEGIN,
+});
+
+export const saveProfileSuccess = () => ({
+  type: SAVE_PROFILE.SUCCESS,
+});
+
+export const saveProfileReset = () => ({
+  type: SAVE_PROFILE.RESET,
+});
+
+export const saveProfileFailure = error => ({
+  type: SAVE_PROFILE.FAILURE,
+  payload: { error },
+});
+
+export const saveProfile = (username, userAccountState, fieldName) => ({
+  type: SAVE_PROFILE.BASE,
   payload: {
     fieldName,
     username,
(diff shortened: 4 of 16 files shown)


config/webpack.common.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   },
   output: {
     path: path.resolve(__dirname, '../dist'),
+    publicPath: '/',
```
davidjoy: Without this, our main.js file gets loaded from the wrong place.

config/webpack.dev.config.js
```diff
@@ -24,6 +25,10 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
+          path.resolve(__dirname, 'node_modules/camelcase-keys'),
```
davidjoy: These are ES6 packages that we depend on now, so they get included in babel.  This is the same fix that frontend-auth got.

config/webpack.prod.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
+          path.resolve(__dirname, 'node_modules/camelcase-keys'),
```
davidjoy: DRYing this up would be nice, but I couldn't decide how to do it and it didn't seem THAT important yet.

src/actions/profile.test.js
```diff
@@ -3,24 +3,24 @@ import {
...
-  saveUserProfilePhotoFailure,
-  saveUserProfilePhotoReset,
-  saveUserProfilePhoto,
-  DELETE_USER_PROFILE_PHOTO,
-  deleteUserProfilePhotoBegin,
-  deleteUserProfilePhotoSuccess,
-  deleteUserProfilePhotoFailure,
-  deleteUserProfilePhotoReset,
-  deleteUserProfilePhoto,
+  SAVE_PROFILE,
```
davidjoy: Gosh, I should add more tests here, I suppose.  Darn.

src/components/NotFoundPage.jsx
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+
+export default class NotFoundPage extends Component {
+  componentDidMount() {}
+
+  render() {
+    return (
+      <div>
```
davidjoy: Pretty lame ATM.  I stole the language from the main edx.org 404 page.

src/components/NotFoundPage.jsx
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+
+export default class NotFoundPage extends Component {
+  componentDidMount() {}
```
davidjoy: whoops.  This could be a functional component.  Didn't put much thought into it.

src/components/UserProfile/index.jsx
```diff
@@ -33,6 +32,10 @@ class UserProfile extends React.Component {
     this.onVisibilityChange = this.onVisibilityChange.bind(this);
   }
 
+  componentDidMount() {
```
davidjoy: So once we load the actual "profile page" part of our app, we then know what the route is and can try to load the user profile for the given user.  

Now that I think about it, we'll need some sort of redirect-based error handling if we're unable to load the profile (bad username, for instance).  This handling should go in the saga, since it catches errors from the backend.

abutterworth:
Looks good to me!

src/components/UserProfile/index.jsx
```diff
@@ -195,11 +198,17 @@ UserProfile.propTypes = {
...
-  saveUserProfile: PropTypes.func,
-  saveUserProfilePhoto: PropTypes.func.isRequired,
-  deleteUserProfilePhoto: PropTypes.func.isRequired,
+  fetchProfile: PropTypes.func.isRequired,
+  saveProfile: PropTypes.func.isRequired,
+  saveProfilePhoto: PropTypes.func.isRequired,
+  deleteProfilePhoto: PropTypes.func.isRequired,
   openEditableField: PropTypes.func.isRequired,
   closeEditableField: PropTypes.func.isRequired,
+  match: PropTypes.shape({
```
davidjoy: This is stuff from react-router.

src/containers/UserProfile/index.jsx
```diff
@@ -2,32 +2,43 @@ import { connect } from 'react-redux';
...
-  username: state.userAccount.username,
-  userLocation: state.userAccount.country,
-  education: state.userAccount.levelOfEducation,
-  socialLinks: state.userAccount.socialLinks,
-  bio: state.userAccount.bio,
-  certificates: null,
-});
+const mapStateToProps = (state) => {
+  const profileImage =
+    state.profilePage.profile.profileImage != null
```
davidjoy: This was blowing up for me without this conditional check, since profileImage didn't exist in the reducer by default.  I think this is a better way of handling it than having a sparsely populated skeleton in the reducer - I recall being bitten by trying it that way in the past...

src/data/reducers/ProfilePageReducer.js
```diff
@@ -1,21 +1,21 @@
...
+  FETCH_PROFILE,
 } from '../../actions/profile';
 
-
 const initialState = {
   error: null,
   saveState: null,
   savePhotoState: null,
   currentlyEditingField: null,
+  profile: {},
```
davidjoy: Renamed this reducer to "profilePage" so that this part of its data is `profilePage.profile` instead of `profile.profile`.  Seems more accurate to what the reducer handles, too.

config/webpack.dev.config.js
abutterworth: I think I want to get rid of these dependencies by writing out the key, key map. For example: 
```{
   levelOfEducation: 'level_of_education'
}```

This could eliminate the need for snakecase and camelcase transforms as well as make our code explicit about what exists.

File: src/data/store.js [hunk 11]
```diff
@@ -18,9 +17,6 @@ const store = createStore(
   composeWithDevTools(applyMiddleware(thunkMiddleware, sagaMiddleware, loggerMiddleware)),
 );
 
-const apiService = new UserAccountApiService(apiClient, process.env.LMS_BASE_URL);
```
davidjoy: Since we don't use UserAccountApiService in the sagas now, this can all go away.

src/sagas/RootSaga.js
```diff
@@ -1,138 +1,110 @@
...
-let userAccountApiService = null;
-
-const PROP_TO_STATE_MAP = {
-  fullName: 'name',
-  userLocation: 'country',
-  education: 'levelOfEducation',
-  socialLinks: socialLinks => socialLinks.filter(({ socialLink }) => socialLink !== null),
-};
-
-export const mapDataForRequest = (props) => {
```
davidjoy: This gets moved into the ProfileApiService.

src/sagas/RootSaga.js
```diff
@@ -1,138 +1,110 @@
...
     // Tells the profile form that
-    yield put(saveUserProfileSuccess());
-    // TODO: export the fetchUserAccountSuccess action from frontend-auth so we can
-    // dry this up.
-    yield put({
-      type: 'FETCH_USER_ACCOUNT_SUCCESS',
-      payload: { userAccount },
-    });
+    yield put(saveProfileSuccess());
+    yield put(fetchProfileSuccess(profile));
```
davidjoy: This is no longer a load of the user account - it's a load of the profile, so we can remove the dependency here on anything from frontend-auth, which feels nice.

src/sagas/RootSaga.js
```diff
@@ -1,138 +1,110 @@
...
+export function* handleSaveProfilePhoto(action) {
   const { username, formData } = action.payload;
 
   try {
-    yield put(saveUserProfilePhotoBegin());
+    yield put(saveProfilePhotoBegin());
 
     // Use call([context, fnName], ...args) for proper context on apiService
-    yield call([userAccountApiService, 'saveUserProfilePhoto'], username, formData);
+    yield call(ProfileApiService.postProfilePhoto, username, formData);
```
davidjoy: No need to do any binding on these anymore.

src/sagas/RootSaga.js
```diff
@@ -1,138 +1,110 @@
...
 
     // Get the account data. Saving doesn't return anything on success.
-    const userAccount = yield call(
-      [userAccountApiService, 'getUserAccount'],
-      username,
-    );
+    yield handleFetchProfile(fetchProfileAction);
```
davidjoy: This can delegate this part to the whole fetch profile saga, which is nice and DRYer.

davidjoy:
Note that this does NOT update the UI to give the user the proper experience when viewing someone else's profile.

albemarle:
One thing to consider for the future-- it can be nice to do refactors & feature additions in separate PRs, so each one has only one thing it's doing.

src/actions/profile.js
```diff
@@ -2,9 +2,10 @@ import AsyncActionType from './AsyncActionType';
 
 export const EDITABLE_FIELD_OPEN = 'EDITABLE_FIELD_OPEN';
 export const EDITABLE_FIELD_CLOSE = 'EDITABLE_FIELD_CLOSE';
-export const SAVE_USER_PROFILE = new AsyncActionType('PROFILE', 'SAVE_USER_PROFILE');
-export const SAVE_USER_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'SAVE_USER_PROFILE_PHOTO');
-export const DELETE_USER_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'DELETE_USER_PROFILE_PHOTO');
+export const FETCH_PROFILE = new AsyncActionType('PROFILE', 'FETCH_PROFILE');
+export const SAVE_PROFILE = new AsyncActionType('PROFILE', 'SAVE_PROFILE');
+export const SAVE_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'SAVE_PROFILE_PHOTO');
+export const DELETE_PROFILE_PHOTO = new AsyncActionType('PROFILE', 'DELETE_PROFILE_PHOTO');
```
albemarle: I like the rename.

src/components/NotFoundPage.jsx
albemarle: Pretty sure you also promised cats.

config/webpack.dev.config.js
davidjoy: I'm totally down with that.

src/components/NotFoundPage.jsx
davidjoy: Crap!

src/sagas/RootSaga.js
```diff
@@ -1,138 +1,110 @@
...
+  try {
+    yield put(saveProfileBegin());
+    const profile = yield call(
+      ProfileApiService.patchProfile,
       username,
-      mapDataForRequest(userAccountState),
+      userAccountState,
     );
 
     // Tells the profile form that
```
albemarle: The suspense is killing me!

davidjoy:
@albemarle Totally agreed on the refactors in separate PRs... I intended do that, actually, but then realized that a chunk of the code in my refactor depended on the actual feature that I'd stashed.  So, er, I just popped the stash and put 'em both in here rather than undo work I'd already done. :)

src/sagas/RootSaga.js
davidjoy: Hah!  Whoops.  Will remove.

src/actions/profile.js
davidjoy: I like simple! :D

PR #24, opened 2019-02-20
Properly display fields on other users profiles.

davidjoy:
If a field isn’t present in another users profile (they either haven’t set it, or it’s not publicly visible), then hide it completely.  If the field is present, then show it in a static way.

The change to EmptyContent was because I kept getting a warning from it, and I think `node` is more accurate to what we expect there.  Similarly, I modified TransitionReplace so that it could render "nothing at all" if desired.  This is used for hidden fields as described above.

### DIFF
--- a/src/components/UserProfile/elements/EmptyContent.jsx
+++ b/src/components/UserProfile/elements/EmptyContent.jsx
@@ -45,7 +45,7 @@ export default EmptyContent;
 
 EmptyContent.propTypes = {
   onClick: PropTypes.func,
-  children: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
+  children: PropTypes.node,
   showPlusIcon: PropTypes.bool,
 };
 
--- a/src/components/UserProfile/index.jsx
+++ b/src/components/UserProfile/index.jsx
@@ -88,6 +88,7 @@ class UserProfile extends React.Component {
       education,
       socialLinks,
       certificates,
+      isCurrentUserProfile,
     } = this.props;
 
     const commonProps = {
@@ -101,8 +102,26 @@ class UserProfile extends React.Component {
     };
 
     const getMode = (name) => {
-      if (name === this.props.currentlyEditingField) return 'editing';
-      if (!this.props[name] || !this.props[name].length) return 'empty';
+      // If the prop doesn't exist, that means it hasn't been set (for the current user's profile)
+      // or is being hidden from us (for other users' profiles)
+      const propExists = this.props[name] != null && this.props[name].length > 0;
+
+      // If this isn't the current user's profile...
+      if (!isCurrentUserProfile) {
+        // then there are only two options: static or nothing.
+        // We use 'null' as a return value because the consumers of
+        // getMode render nothing at all on a mode of null.
+        return propExists ? 'static' : null;
+      }
+      // Otherwise, if this is the current user's profile...
+      if (name === this.props.currentlyEditingField) {
+        return 'editing';
+      }
+
+      if (!propExists) {
+        return 'empty';
+      }
+
       return 'editable';
     };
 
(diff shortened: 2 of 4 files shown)


src/components/UserProfile/index.jsx
```diff
@@ -101,8 +102,25 @@ class UserProfile extends React.Component {
...
     const getMode = (name) => {
-      if (name === this.props.currentlyEditingField) return 'editing';
-      if (!this.props[name] || !this.props[name].length) return 'empty';
+      // If the prop doesn't exist, that means it hasn't been set (for the current user's profile)
+      // or is being hidden from us (for other users' profiles)
+      const propExists = this.props[name] != null && this.props[name].length > 0;
+      // If this isn't the current user's profile, then there are only two options:
+      // static or nothing. We use 'null' as a return value because the consumers of
+      // getMode render nothing at all on a mode of null.
+      if (!isCurrentUserProfile) {
```
albemarle: I'm pondering if viewingOwnProfile or something similar would work?  isCurrentUserProfile makes me wonder if there are two profiles in play, one is the current one and one is the previous profile.  But it's also entirely possible that that's just me.  :-)

src/components/UserProfile/index.jsx
```diff
@@ -101,8 +102,25 @@ class UserProfile extends React.Component {
...
+        return propExists ? 'static' : null;
+      }
+      // Otherwise, if this is the current user's profile and it's name matches the
```
albemarle: You can probably leave out this comment and the two below it, since the code is pretty clear.  If you do keep them, small typo, it's -> its.

src/components/UserProfile/index.jsx
davidjoy: Ah - well, just depends on our language for who the active session is.  I've tended to always call that the "current user" or the "active user" or something like that, so that's "is the profile of the current user", rather than "is the current profile that we're viewing". :) If we have some guidelines there I'm more than happy to line up with em!

src/components/UserProfile/index.jsx
davidjoy: I have this vague philosophy that as soon as a method gets too complex to figure out what's going on at a glance, it's time to comment the hell out of it. ;) 

But yes, fair.  That part is pretty self-explanatory.

For years I implicitly understood its vs. it's.  Somehow in the past month that appears to have gone out the window. :)

albemarle: Depends on what kind of complexity.  In general comments should explain *why* you're doing something.  If what you need to explain is *what* you're doing, I'd reach first for well named intermediary variables, well named methods, and other ways to communicate through the code itself.  Comments are sort of worse than tests, in that when they become out of date or wrong, they don't even break to let you know!

albemarle: That said, yeah, sometimes you're just stuck with a long list of things you need to do in order and comments can at least signpost the main chunks.  I don't think this method has hit that point though-- it's still young and spry.  ;-)

src/components/UserProfile/elements/EmptyContent.jsx
```diff
@@ -45,7 +45,7 @@ export default EmptyContent;
 
 EmptyContent.propTypes = {
   onClick: PropTypes.func,
-  children: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
+  children: PropTypes.node,
```
abutterworth: Good stuff, I didn't know this existed. From the docs: 
```
// Anything that can be rendered: numbers, strings, elements or an array
// (or fragment) containing these types.
optionalNode: PropTypes.node,
```

src/components/UserProfile/index.jsx
```diff
@@ -101,8 +102,26 @@ class UserProfile extends React.Component {
     };
 
     const getMode = (name) => {
-      if (name === this.props.currentlyEditingField) return 'editing';
-      if (!this.props[name] || !this.props[name].length) return 'empty';
+      // If the prop doesn't exist, that means it hasn't been set (for the current user's profile)
+      // or is being hidden from us (for other users' profiles)
+      const propExists = this.props[name] != null && this.props[name].length > 0;
```
abutterworth: Should this be `!== null` ?

PR #27, opened 2019-02-22
WIP - using i18next instead of react-intl - for evaluation only

albemarle:


albemarle:
Closing as we didn't find a format that Transifex can use.

PR #29, opened 2019-02-22
TransitionReplace Improvement

abutterworth:
Since we've added the ability for transition replace to have no content, we can't guarantee an onChildExited event will fire resetting the hieght to null

### DIFF
--- a/src/components/UserProfile/elements/TransitionReplace.jsx
+++ b/src/components/UserProfile/elements/TransitionReplace.jsx
@@ -36,20 +36,21 @@ class TransitionReplace extends React.Component {
   }
 
   onChildEntering(htmlNode) {
-    // subtract 1 from the height to improve shifts at the end of the animation
     this.setState({
-      height: htmlNode.offsetHeight - 1,
+      height: htmlNode.offsetHeight,
     });
   }
 
   onChildEntered(htmlNode) {
     if (this.props.onChildEntered) this.props.onChildEntered(htmlNode);
+    this.setState({
+      height: null,
+    });
   }
 
   onChildExit(htmlNode) {
-    // subtract 1 from the height to improve shifts at the end of the animation
     this.setState({
-      height: htmlNode.offsetHeight - 1,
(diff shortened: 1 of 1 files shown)


davidjoy:
Not sure I fully understand what it's fixin' or how it's fixin' it...

src/components/UserProfile/elements/TransitionReplace.jsx
```diff
@@ -36,20 +36,21 @@ class TransitionReplace extends React.Component {
...
-      height: htmlNode.offsetHeight - 1,
+      height: htmlNode.offsetHeight,
     });
   }
 
   onChildEntered(htmlNode) {
     if (this.props.onChildEntered) this.props.onChildEntered(htmlNode);
+    this.setState({
+      height: null,
+    });
```
abutterworth: By setting the height to `null` here we are unsetting the height (which should always be null when there is no transition happening) after a child has finished entering. This is necessary now that we can't rely on a childExited event, given that the content of this component can be empty.

src/components/UserProfile/elements/TransitionReplace.jsx
```diff
@@ -36,20 +36,21 @@ class TransitionReplace extends React.Component {
   }
 
   onChildEntering(htmlNode) {
-    // subtract 1 from the height to improve shifts at the end of the animation
     this.setState({
-      height: htmlNode.offsetHeight - 1,
+      height: htmlNode.offsetHeight,
```
abutterworth: This change and the identical one just removes a small hack that doesn't seem to be doing anything anymore.

src/components/UserProfile/elements/TransitionReplace.jsx
albemarle: Is the above worth adding as a comment in the code?

abutterworth: The statement applies to the component as a whole, not just that line. Do you think it's worth describing how the component works at the top of the file?

PR #30, opened 2019-02-22
Update empty states to match UX recommendation

abutterworth:
They look like links now

### DIFF
--- a/src/components/UserProfile/elements/EmptyContent.jsx
+++ b/src/components/UserProfile/elements/EmptyContent.jsx
@@ -1,18 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+import { Button } from 'reactstrap';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
 
 function EmptyContent({ children, onClick, showPlusIcon }) {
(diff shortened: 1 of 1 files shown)


src/components/UserProfile/elements/EmptyContent.jsx
```diff
@@ -1,41 +1,32 @@
...
-  }
-
   return (
-    <div {...props}>
-      {showPlusIcon ? <FontAwesomeIcon className="ml-1 mr-3" icon={faPlus} /> : null}
-      {children}
+    <div>
+      {onClick ? (
+        <Button
+          color="link"
```
davidjoy: Buttons are supposed to have a `type` attribute, right?  Think the a11y linter complains about that...

davidjoy: Maybe reactstrap makes it `type="button"` by default, I suppose.

abutterworth: The source for Button https://github.com/reactstrap/reactstrap/blob/master/src/Button.js

Surprisingly the type attribute doesn't seem to be addressed at all there, though if specified it will pass through. I'll swap this with the Paragon Button

robrap: If reactstrap is our first choice to avoid maintenance, any use in trying to open a PR against reactstrap to try to address issues like this? (This proposal would be in parallel, in addition to switching to Paragon for solving today's issue.)

PR #31, opened 2019-02-22
WIP - demo only - react-intl example of using just a string

albemarle:


src/components/UserProfile/MyCertificates.jsx
```diff
@@ -92,11 +93,11 @@ function MyCertificates({
...
-            <FormattedMessage
-              id="profile.no.certificates"
-              defaultMessage="You don't have any certificates yet."
-              description="displays when user has no course completion certificates"
-            />
+            {intl.formatMessage({
+              id: 'profile.no.certificates',
+              defaultMessage: 'You don\'t have any certificates yet.',
+              description: 'displays when user has no course completion certificates',
+            })}
```
abutterworth: Good to see this out of a React component, but it still seem quite verbose. Are we oppose to something terse like:
`t('You don\'t have any certificates yet.')`

I know we lose the id and the description... Curious on your thoughts here.

PR #33, opened 2019-02-25
First pass - realigning/simplifying/refactoring

davidjoy:
# Background on this PR

Last week, things started getting a bit difficult to reason as @abutterworth and I were working on getting all the forms working.  

Initial states, dirty states, pristine, usage of component state at various levels, difficulty in merging redux state into component state in an understandable way, all came to a head.  Functioning of the application relied partially on subtle react virtual DOM rules about when components get remounted, and we didn't have a consistent, understandable way of pulling profile state out of redux and working with it.  

This was compounded by the dual nature of our server calls - separate calls for saving profile data and saving preferences.  We realized that the application itself should not care how the data gets sent over the wire for the most part - conceptually we have some simple forms that should be easy to manage, and so we set out to re-align the code with that fact in mind.

This PR is a precursor to one in which we'll fully align on that vision.  In the meantime, there were a number of features of our application structure that I found were slowing me down, adding friction.  I also saw a few inconsistencies in our directory structure and decided to just do a pass of aligning things a bit above and beyond what was strictly necessary.

# Changes

## Miscellaneous small renames

You'll find some renamed constants and functions, just for simplicity's sake.  

## Combining “containers” into “components”.

A component's propTypes, defaultProps, and mapStateToProps function are intimately tied together, and all need to align.  Having the component's `connect()` call in a separate file didn't really help much, and just required context switching back and forth to tell what was going on.  Furthermore, as an app gets more complex, there may be a number of connected components at various levels, and maintaining a dual directory structure just becomes a bit cumbersome.  It's likely that we end up connecting some of our individual form components, so this seemed apropos to get out of the way.

## Flatten out "data" into "reducers" and "config"

I've always found it helpful in a react/redux application to think of reducers, components, actions, sagas/thunks, and services as peers that work together and help you find the right place for any given piece of code. Furthermore, there's some configuration-like code that an application uses at startup. 
 This PR flattens out “data” into “reducers” and “config” to consolidate configuration-like files in one place and to make reducers a peer of its teammates (components, actions, sagas, and services).

## Creating dev/prod-specific redux configurations.

We're going to want dev/prod-specific store configurations.  The prod one shouldn't include redux dev tools or the logger.  The structure ends up looking very similar to the webpack configuration.

## Converting “index.jsx” files into files named for their contents.  

This is merely a quality-of-life improvement, and one that was a constant source of friction. We had four different "index.js(x)" files in the app.  This gives them names appropriate to their contents, reducing keystrokes and helping to differentiate tabs in IDEs.

## Splitting up the top-level “index.jsx” file into an entry point and an “App” component.

This is a separation of concerns change - the top-level index.jsx file is the entry point to the app and the place where we mount our component hierarchy on the DOM and setup our store.  The new file "App.jsx", takes the guts out of the entry point and puts the component portion of it into a separate file in the components directory, closer to where it belongs.

## Renaming SCSS file to “index.scss” to keep it consistent with where it’s imported.

Simple change to keep it consistent.  Also, SASS can have issues importing multiple files at multiple levels, so it's a somewhat good idea not to put in peoples' heads that they should have a SASS file per component unless we really want to go that way.


src/actions/ProfileActions.js
```diff
@@ -57,12 +70,13 @@ export const saveProfileFailure = error => ({
...
 });
 
-export const saveProfile = (username, userAccountState, fieldName) => ({
+export const saveProfile = (username, { profileData, preferencesData }, fieldName) => ({
   type: SAVE_PROFILE.BASE,
   payload: {
     fieldName,
     username,
-    userAccountState,
+    profileData,
```
davidjoy: So this structure currently _works_ but is going to be refactored to be simpler.  Rather than passing around two separate structures for data, it'll all just be in one place as far as most of the application is concerned.

src/components/App.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+import HeaderLogo from '../../assets/edx-sm.png';
+import FooterLogo from '../../assets/edx-footer.png';
+import NotFoundPage from './NotFoundPage';
+
+class App extends Component {
+  componentDidMount() {
+    const { username } = this.props;
+    const userAccountApiService = new UserAccountApiService(apiClient, process.env.LMS_BASE_URL);
+    this.props.fetchUserAccount(userAccountApiService, username);
```
davidjoy: This has been refactored to just be connected to redux, that way it can get `authentication` state and its `fetchUserAccount` action the "right" way.

src/components/ProfilePage.jsx
```diff
@@ -38,45 +49,32 @@ class UserProfile extends React.Component {
...
-    if (valueToSave != null) {
-      this.props.saveProfile(
-        this.props.username,
-        {
-          [fieldName]: valueToSave,
-        },
-        fieldName,
-      );
-    }
+    const data = {};
```
davidjoy: This `onSave` code is likely to be further refactored and simplified - again, it currently _works_, but I wouldn't critique it too much in this PR.

src/components/ProfilePage.jsx
```diff
@@ -1,16 +1,28 @@
...
+
+// Actions
+import {
+  fetchProfile,
+  saveProfile,
+  saveProfilePhoto,
+  deleteProfilePhoto,
+  openField,
+  closeField,
+} from '../actions/ProfileActions';
```
abutterworth: I'm not particularly happy with combining this Page component with the higher order redux-connected component. I see actions imported at the top that are later mapped (with dispatch) to props. It's tempting to just use those imported functions since they're the same name.

I think we would be better served to keep these things separate, though IMO ideally right next to each other in the file structure. Something like: `ProfilePage.jsx` and `ProfilePageConnected.jsx`

davidjoy: Yeaaaah, lemme think about that for a sec... it's a good point.  Normally I just universally do `this.props.fetchProfile` (or whatever action) when I need it, which avoids the problem, but I can see how that'd be confusing.  🤔

PR #34, opened 2019-02-26
David/small fix and parallelizing

davidjoy:
Didn't feel like making two separate branches.

1. Fixes an error in how platform names are given to StaticListItem.
2. Parallelizes fetch profile calls.

### DIFF
--- a/src/components/ProfilePage/SocialLinks.jsx
+++ b/src/components/ProfilePage/SocialLinks.jsx
@@ -15,7 +15,6 @@ const brandIcons = {
   linkedin: faLinkedin,
 };
 
-
 class SocialLinks extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


src/components/ProfilePage/SocialLinks.jsx
```diff
@@ -26,10 +25,12 @@ class SocialLinks extends React.Component {
   }
 
   onSave() {
-    const values = this.props.platforms.filter(({ key }) => typeof this.state[key] !== 'undefined').map(({ key }) => ({
-      platform: key,
-      socialLink: this.state[key],
-    }));
+    const values = this.props.platforms
+      .filter(({ key }) => typeof this.state[key] !== 'undefined')
```
albemarle: Could also be !_.isUndefined(this.state[key])

(The lack of _.isDefined() is a perpetual sorrow to me.)

src/components/ProfilePage/SocialLinks.jsx
```diff
@@ -190,16 +192,15 @@ SocialLink.propTypes = {
...
-  platform,
-  onClickEmptyContent,
-  name,
+  url, platform, onClickEmptyContent, name,
 }) {
-  const linkDisplay = url != null ?
-    <SocialLink name={name} url={url} platform={platform} /> :
-    <EmptyContent onClick={onClickEmptyContent}>Add {name}</EmptyContent>;
+  const linkDisplay =
+    url != null ? (
```
albemarle: Philosophical note about two things in tension here: the case where it's not null is the one you care more about, so it makes sense to put it first, but on the other hand, conditionals are easiest for humans to read if they're not unpositive.  I tend to prioritize the second consideration more than the first.

abutterworth:
Heads up I have a reworked version of social links in a different PR.

davidjoy:
This PR has been subsumed by #40.  Closing.

PR #37, opened 2019-02-26
Add age messaging

abutterworth:
Looks only at the `requiresParentalConsent` boolean. It seems that this is false when `yearOfBirth` is null or less than 13 years ago. It's worth noting that this differs from the [current logic on the profile page](https://github.com/edx/edx-platform/blob/1a8df4517943a15b2d1c71be8ea44c4f6382d00c/lms/static/js/student_account/models/user_account_model.js#L58) seen here:
```
isAboveMinimumAge: function() {
    var yearOfBirth = this.get('year_of_birth');
    var isBirthDefined = !(_.isUndefined(yearOfBirth) || _.isNull(yearOfBirth));
    return isBirthDefined && !(this.get('requires_parental_consent'));
}
```

### DIFF
--- a/src/components/ProfilePage.jsx
+++ b/src/components/ProfilePage.jsx
@@ -21,6 +21,7 @@ import Education from './ProfilePage/Education';
 import SocialLinks from './ProfilePage/SocialLinks';
 import Bio from './ProfilePage/Bio';
 import MyCertificates from './ProfilePage/MyCertificates';
+import AgeMessage from './ProfilePage/AgeMessage';
 
 export class ProfilePage extends React.Component {
   constructor(props) {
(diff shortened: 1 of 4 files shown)


src/components/ProfilePage.jsx
```diff
@@ -210,6 +211,8 @@ export class ProfilePage extends React.Component {
             </Col>
             <Col xs={{ order: 1 }} md={{ size: 8, order: 2 }} lg={{ size: 8, offset: 1 }} className="mt-4 mt-md-n5">
 
+              {this.props.requiresParentalConsent ? <AgeMessage accountURL="#account" /> : null}
```
douglashall: Shouldn't this link to the LMS account settings page until we create the new microfrontend version of that page?

PR #38, opened 2019-02-26
Add event log call and new event.

robrap:
This is ready outside of tests.


robrap:
FYI: @edx/arch-squad: I've got some questions in my code comments I want to discuss with someone.  Thanks.

package-lock.json
```diff
@@ -9281,8 +9333,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
```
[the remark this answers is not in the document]
robrap: I'm just going to ignore this and hope all is well.

src/services/NewRelicService.js
```diff
@@ -0,0 +1,27 @@
...
+  static logAPIErrorResponse(error) {
+    let { message } = error;
+    if (error.response) {
+      message = `${error.response.status} ${error.response.config.url} ${JSON.stringify(error.response.data)}`;
+    } else if (error.request) {
+      message = `${error.request.status} ${error.request.responseURL} ${error.request.responseText}`;
+    } else if (error.stack) {
+      message = error.stack;
+    }
+    this.logError(new Error(`API request failed: ${message}`));
```
abutterworth: `this` is unneeded here?

src/services/NewRelicService.js
```diff
@@ -0,0 +1,27 @@
+class NewRelicService {
```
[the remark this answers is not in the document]
abutterworth: I think this is fine to leave in each project. If we do move it out I would suggest we change the name to something like LoggingService to abstract away New Relic. And then flesh out an API to do everything we need.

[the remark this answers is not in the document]
robrap: Even here I could rename it.  I will do that.

src/services/NewRelicService.js
robrap: @abutterworth: Linter failed saying logError was undefined when I tried to remove this.

src/analytics.js
```diff
@@ -1,11 +1,30 @@
...
+// Sends events to Segment and downstream
 const handleTrackEvents = (eventName, properties) => {
   // Simply forward track events to Segment
   window.analytics.track(eventName, properties);
 };
 
 
+// Sends events to tracking log and downstream
+// TODO: Determine consistent naming for eventName vs eventType and properties v eventData.
+const logEvent = (eventType, eventData) => {
```
abutterworth: Do we have an opinion on using this syntax vs below in this context?
```
function logEvent(eventType, eventData) {

}
```

robrap: I think you do, so I changed it. :)

abutterworth: Haha, I don't know if there's a proper way or not

robrap: The linter should state our proper way, and it is happy with both.  I like the functions more, and you seemed like you might, so I went for it.

robrap: @douglashall? @davidjoy?

PR #39, opened 2019-02-27
fix(certificates): use real certificate data

douglashall:
This PR depends on https://github.com/edx/edx-platform/pull/19884 which adds a certificate list API endpoint on the backend.

### DIFF
--- a/src/services/ProfileApiService.js
+++ b/src/services/ProfileApiService.js
@@ -1,4 +1,7 @@
+import camelcaseKeys from 'camelcase-keys';
+
 import apiClient from '../config/apiClient';
+import CERTIFICATE_TYPES from '../constants/certificates';
 import { configuration } from '../config/environment';
 import { unflattenAndTransformKeys, flattenAndTransformKeys } from './utils';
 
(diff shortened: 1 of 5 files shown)


src/services/ProfileApiService.js
```diff
@@ -131,30 +132,39 @@ export function patchPreferences(username, preferences) {
...
+    //       name: 'Micro Masters',
+    //     },
+    //     title: 'Kazoo Pinch Harmonics',
+    //     organization: 'New England Conservatory',
+    //     downloadUrl: 'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcR88selLYRHDQemO8MnY3w3ajlSuKZcoRhVTp3eGvKq2HMRTn8q',
+    //   },
+    // ];
+    // setTimeout(() => {
+    //   resolve(dummyData);
+    // }, 200);
```
albemarle: Are we ready to delete the fake data?

davidjoy: Given that the format of it isn't valid anymore, probably.

davidjoy:
Is there a companion PR to this in edx-platform that we should get in too?

douglashall:
@davidjoy Can you give this one another look now that it is ready for review?

PR #35, opened 2019-02-26
Form errors

abutterworth:
- Each field now utilizes a form element and captures onSubmit events instead of onClick events from a save button.
- Errors on each field driven by props supplied to the ProfilePage

PR #40, opened 2019-02-27
Re-re-rewire data/form state up to redux

davidjoy:
This PR rewires how our forms get their data.  It also folds in changes from #35 because @abutterworth and I had a pretty hefty conflict between our two branches.

This turned into a broader effort than I intended, admittedly.  

The original goal was to take our form state out of component state and put it into redux state for all six forms on the profile page.  This has been done, but doing so begged a bit of broader refactoring and renaming as well.

In no particular order:

1. We were referring to our forms as fields - that language has been fixed.  They're forms now.
2. We were putting 'draft' form values in component state - that state has been moved up to the redux level.
3. The fetchProfile action is now responsible for making several underlying calls to the system for the profile's account information, preferences (if it's your own account), and certificates.  Before, we had multiple actions responsible for each part, and coordination was hard.
4. `react-router-redux` has been replaced with `connected-react-router`.  This is admittedly not something that ultimately needed to be in this PR - I _thought_ I was going to need it, but after the work was done, the need for the change fell out and it's essentially unrelated.  That said, `react-router-redux` is deprecated and no longer supported, and `connected-react-router` has taken its place.  I didn't see any reason to throw away the work, so here it is.
5. I updated some packages as part of number 4 above which didn't strictly need to be updated.  I _thought_ I needed to update them w/r/t changing the router, but I didn't.  As above, didn't see any reason to throw away the work, though.
6. Introduced `reselect` to handle derived state.  This helps keep components clean. 
7. Directly connected the forms to the store.  This makes ProfilePage.jsx a lot cleaner.
8. The ProfilePage.jsx file still manages calling action creators to put stuff in redux - the forms are connected for their data, but are otherwise "dumb" about how the data gets back into redux.  It felt weird, for instance, to have the Name.jsx component call an action creator called `saveProfile`... just seemed above its pay grade.  Admittedly this is a bit asymmetrical, but it allowed ProfilePage.jsx to be a lot shorter/have less responsibility for passing data down.


package.json
```diff
@@ -39,6 +39,7 @@
     "bootstrap": "^4.2.1",
     "camelcase-keys": "^5.0.0",
     "classnames": "^2.2.5",
+    "connected-react-router": "^5.0.1",
```
davidjoy: As said in the summary, changing the router and upgrading the react/redux packages wasn't strictly necessary, but I didn't see any reason to waste the work.

src/actions/ProfileActions.js
```diff
@@ -1,42 +1,31 @@
...
+  type: FETCH_PROFILE.BASE,
+  payload: { username },
 });
 
 export const fetchProfileBegin = () => ({
   type: FETCH_PROFILE.BEGIN,
 });
 
-export const fetchProfileSuccess = profile => ({
+export const fetchProfileSuccess = (account, preferences, certificates) => ({
```
davidjoy: fetchProfileSuccess dumps all our fancy new data into the reducer now - not just the misnomered "profile" (which is really "account" data)

src/actions/ProfileActions.js
```diff
@@ -48,17 +37,25 @@ export const fetchProfileReset = () => ({
   type: FETCH_PROFILE.RESET,
 });
 
-export const fetchProfile = username => ({
-  type: FETCH_PROFILE.BASE,
-  payload: { username },
+// SAVE PROFILE ACTIONS
+
+export const saveProfile = formId => ({
```
davidjoy: To save a profile, we just need to pass which form was being submitted.  The Saga can get the full data out of the reducer itself.

src/components/ProfilePage.jsx
```diff
@@ -9,152 +9,78 @@ import {
...
 import SocialLinks from './ProfilePage/SocialLinks';
 import Bio from './ProfilePage/Bio';
-import MyCertificates from './ProfilePage/MyCertificates';
+import Certificates from './ProfilePage/Certificates';
 
 export class ProfilePage extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = {
```
davidjoy: Goodnight, sweet state-prince.

src/components/ProfilePage.jsx
```diff
@@ -9,152 +9,78 @@ import {
...
+  updateVisibilityDraft,
+  updateAccountDraft,
 } from '../actions/ProfileActions';
 
 // Components
 import ProfileAvatar from './ProfilePage/ProfileAvatar';
-import FullName from './ProfilePage/FullName';
-import UserLocation from './ProfilePage/UserLocation';
+import Name from './ProfilePage/Name';
+import Country from './ProfilePage/Country';
```
davidjoy: I aligned some names with what the server sends.

src/components/ProfilePage.jsx
```diff
@@ -9,152 +9,78 @@ import {
...
-
-  onVisibilityChange(fieldName, visibility) {
-    this.setState({
-      [fieldName]: {
-        value: this.state[fieldName].value,
-        visibility,
-      },
-    });
+  handleChange(formId, name, value) {
+    if (name === 'visibility') {
```
davidjoy: Depending on what kind of field changed, we store the data in different places in the reducer.  Separate actions felt like a relatively clean way to tell the difference.

src/components/ProfilePage.jsx
albemarle: 🤴 💀 💤

src/sagas/RootSaga.js
```diff
@@ -24,93 +24,75 @@ import {
...
+    }
+
+    const result = yield all(calls);
+
+    if (result.length > 2) {
+      const [account, certificates, preferences] = result;
+      yield put(fetchProfileSuccess(account, preferences, certificates));
+    } else {
+      const [account, certificates] = result;
+      yield put(fetchProfileSuccess(account, { visibility: {} }, certificates));
```
abutterworth: This may be a little weird, recognize the need tho

src/sagas/RootSaga.js
```diff
@@ -24,93 +24,75 @@ import {
...
-      responseData.profile = yield call(
-        ProfileApiService.patchProfile,
-        username,
-        profileData,
-      );
+    let accountResult = null;
+    // Build the visibility drafts into a structure the API expects.
+    const preferences = {
+      visibility: visibilityDrafts,
+    };
```
abutterworth: Let's keep in mind that we need to afford for other preferences. For example:
```
{
   accountPrivacy: 'custom',
   visibility: { /* field visibilities */ },
}

File: src/selectors/ProfilePageSelector.js [hunk 9]
```diff
@@ -0,0 +1,119 @@
...
+  (username, profileAccount) => username === profileAccount.username,
+);
+
+export const editableFormModeSelector = createSelector(
+  profileAccountSelector,
+  profileCertificatesSelector,
+  formIdSelector,
+  isCurrentUserProfileSelector,
+  currentlyEditingFieldSelector,
+  (account, certificates, formId, isCurrentUserProfile, currentlyEditingField) => {
```
abutterworth: FWIW I'm not a huge fan of this composable selector pattern

src/actions/ProfileActions.js
```diff
@@ -1,42 +1,31 @@
...
-  type: FIELD_CLOSE,
-  fieldName,
-});
-
-export const updateDrafts = drafts => ({
-  type: UPDATE_DRAFTS,
-  drafts,
+export const fetchProfile = username => ({
+  type: FETCH_PROFILE.BASE,
+  payload: { username },
```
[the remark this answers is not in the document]
davidjoy: The payload thing is a "flux standard action" convention.  It's supposed to be the canonical way an action is structured.  That said, we don't seem to follow the spec completely w/r/t errors:

https://github.com/redux-utilities/flux-standard-action

src/sagas/RootSaga.js
davidjoy: Yeah, I couldn't figure out a more elegant way to handle it.

src/selectors/ProfilePageSelector.js
davidjoy: It's a bit weird to get used to, admittedly.  It really starts shining when you need lots of different kinds of data in lots of different areas of the app...

PR #44, opened 2019-02-28
Add "member since 2017"

abutterworth:
https://openedx.atlassian.net/browse/ARCH-436

The year in "Member since 2017" is now dynamic.

### DIFF
--- a/src/components/ProfilePage/DateJoined.jsx
+++ b/src/components/ProfilePage/DateJoined.jsx
@@ -4,8 +4,7 @@ import PropTypes from 'prop-types';
 function DateJoined({ date }) {
   if (date == null) return null;
 
-  // Not a proper way to do it, but probably cheaper than creating a Date.
-  const year = date.substring(0, 4);
+  const year = (new Date(date)).getFullYear();
 
   return (
     <p className="mb-0">Member since {year}</p>
--- a/src/components/ProfilePage/DateJoined.jsx
+++ b/src/components/ProfilePage/DateJoined.jsx
@@ -1,13 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormattedMessage, FormattedDate } from 'react-intl';
 
 function DateJoined({ date }) {
   if (date == null) return null;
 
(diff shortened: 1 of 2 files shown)


src/components/ProfilePage/DateJoined.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function DateJoined({ date }) {
+  if (date == null) return null;
+
+  const year = (new Date(date)).getFullYear();
+
+  return (
+    <p className="mb-0">Member since {year}</p>
```
albemarle: Moving forward, can we include i18n with all new code?

abutterworth: Will do!

src/components/ProfilePage/DateJoined.jsx
```diff
@@ -1,13 +1,21 @@
...
 
 function DateJoined({ date }) {
   if (date == null) return null;
 
-  const year = (new Date(date)).getFullYear();
-
   return (
-    <p className="mb-0">Member since {year}</p>
+    <FormattedMessage
+      id="profile.date.joined"
```
albemarle: The vague convention I've been using is reponame.componentname.something.about.the.message, so in this case profile.datejoined.member.since is probably what I'd call it.  That said, it actually makes no difference at all.

src/components/ProfilePage/DateJoined.jsx
```diff
@@ -7,7 +7,7 @@ function DateJoined({ date }) {
 
   return (
     <FormattedMessage
-      id="profile.date.joined"
+      id="profile.datejoined.member.since"
```
robrap: @abutterworth @albemarle: When and if code gets refactored into different components, we should make it clear (process-wise) if we would **not** want the keys to also be refactored so that translators don't have to retranslate the same messages.  If that is the case, only the "profile" part of the namespace will probably be reliable longterm.

albemarle: Good point!

PR #45, opened 2019-02-28
Flip buttons and put visibility on new line

abutterworth:
CTAs are left aligned and Primary comes before Secondary


src/components/ProfilePage/elements/VisibilitySelect.messages.jsx
```diff
@@ -0,0 +1,46 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
```
albemarle: Would it make sense to just use the FormControls messages file with a note in the code that they're the same?

src/components/ProfilePage/elements/VisibilitySelect.messages.jsx
```diff
@@ -0,0 +1,46 @@
...
+  'profile.formcontrols.who.can.see': {
+    id: 'profile.formcontrols.who.can.see',
```
albemarle: This duplicate id will (eventually) be flagged by automated tools.

abutterworth: These messages have been removed from the other file. But this should reflect visibility select. Will update

albemarle: At least in this case it's just a search and replace, but I am definitely seeing the drawback of that naming convention.  :-P

PR #47, opened 2019-03-01
Add language proficiency field

abutterworth:
It only displays and saves one – parity with the profile page today.
The API is prepared to handle an array of languages when we're ready

### DIFF
--- a/src/constants/languages.js
+++ b/src/constants/languages.js
@@ -1,0 +1,7 @@
+import LANGUAGES from '@cospired/i18n-iso-languages';
+
+LANGUAGES.registerLocale(require('@cospired/i18n-iso-languages/langs/en.json'));
+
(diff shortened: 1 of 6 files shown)


src/constants/languages.js
```diff
@@ -0,0 +1,7 @@
+import LANGUAGES from '@cospired/i18n-iso-languages';
+
+LANGUAGES.registerLocale(require('@cospired/i18n-iso-languages/langs/en.json'));
+
+const ALL_LANGUAGES = LANGUAGES.getNames('en');
```
albemarle: I'm okay merging this PR, but let's create a followup ticket to i18n this.

abutterworth: All set: ARCH-457

PR #49, opened 2019-03-01
Add version to event.

robrap:
Rather than reverting this event, I'd like to continue to make progress in other areas.  I'm trying out adding versioning info to the event, so I can continue testing in various environments.

### DIFF
--- a/src/components/ProfilePage.jsx
+++ b/src/components/ProfilePage.jsx
@@ -44,6 +44,7 @@ export class ProfilePage extends React.Component {
     this.props.fetchProfile(this.props.match.params.username);
     logEvent('edx.profile.viewed', {
       profileUsername: this.props.match.params.username,
+      version: 'alpha-0.1.0',
     });
   }
 
--- a/src/components/ProfilePage.test.jsx
+++ b/src/components/ProfilePage.test.jsx
@@ -42,6 +42,7 @@ describe('<ProfilePage />', () => {
       expect(analytics.logEvent.mock.calls[0][0]).toEqual('edx.profile.viewed');
       expect(analytics.logEvent.mock.calls[0][1]).toEqual({
         profileUsername: 'test-username',
+        version: 'alpha-0.1.0',
       });
     });
   });


src/components/ProfilePage.jsx
```diff
@@ -44,6 +44,7 @@ export class ProfilePage extends React.Component {
     this.props.fetchProfile(this.props.match.params.username);
     logEvent('edx.profile.viewed', {
```
robrap: Reviewers note: I could also temporarily change to `edx.bi.profile.viewed` in addition to or in place of the `version`, to better match our legacy practices for now.

brianhw: I sympathize with the desire to get a test event out there.  I think we should discuss versioning more before rolling it out, because I would prefer it to be more systematic than ad hoc.  

I prefer the name-change suggestion if this is a place-holder event, and the edx.bi.profile.viewed suggestion as a legacy practice seems good to me.   I would also be in favor of something like "edx.dev.profile.viewed", where we can filter those events from export to researchers based on the prefix "edx.dev" (or a similar convention), but we don't have that in place now and want that to be part of the versioning conversation we have.  (It's not in place for the "edx.bi" events either, but that's more entrenched as a legacy practice. :)

Also, the spec lists the property as "profile_username", but the property below is "profileUsername".  I think we should use the underscore format for greater consistency with the properties of existing events, all else being equal.  That said, I think that just "username" would also be adequate.

robrap:
Thanks @brianhw. 
1. I personally find `profile_username` less confusing, but could just be me.
2. I knew version was rushed, but just wanted to see how it might look. 🙂

Either we’ll wrap up final design before this needs to go to Prod, or I’ll switch to edx.bi for now. Thanks for looking at this.

src/components/ProfilePage.jsx
robrap: FYI: The JavaScript camelcase is automatically converted to snakecase before the event gets sent.

robrap:
Closing. Not using this adhoc versioning scheme.  Using "edx.bi" temporarily.

PR #50, opened 2019-03-01
Update display of avatar upload menu

abutterworth:


src/components/ProfilePage.jsx
```diff
@@ -201,7 +201,7 @@ ProfilePage.defaultProps = {
 const mapStateToProps = (state) => {
   const profileImage =
     state.profilePage.account.profileImage != null
-      ? state.profilePage.account.profileImage.imageUrlLarge
+      ? state.profilePage.account.profileImage.image_url_large
```
douglashall: It would be cool, if we ran `camelcaseKeys` with the `deep: true` option set down at the service layer instead, but this works.

abutterworth: Yea I think David has work that addresses this.

src/components/ProfilePage/ProfileAvatar.jsx
```diff
@@ -32,58 +39,90 @@ class ProfileAvatar extends React.Component {
...
+        {this.props.src == null ? (
+          <Button className="text-white btn-block" color="link" size="sm">
+            <FormattedMessage
+              id="profile.profileavatar.upload-button"
+              defaultMessage="Upload Photo"
+              description="Upload photo button"
             />
-          </form>
+          </Button>
+        ) : (
```
robrap: Nit: Would prefer if/else.

src/components/ProfilePage/ProfileAvatar.jsx
```diff
@@ -32,58 +39,90 @@ class ProfileAvatar extends React.Component {
...
-              innerRef={this.fileInput}
-              type="file"
-              name="file"
-              id="exampleFile"
-              onInput={this.onInput}
-              accept=".jpg, .jpeg, .png"
+  renderMenu() {
+    return (
+      <div className="profile-avatar-menu-container">
+        {this.props.src == null ? (
```
robrap: Nit?: `===` ?

abutterworth:
This doesn't address the viewing other person's profile scenario @douglashall good call.

src/components/ProfilePage/ProfileAvatar.jsx
albemarle: I'm curious if we have linting on for === ...

src/components/ProfilePage.jsx
robrap: Add a comment please.

src/components/ProfilePage/ProfileAvatar.jsx
```diff
@@ -32,58 +39,90 @@ class ProfileAvatar extends React.Component {
...
+              description="Upload photo button"
             />
-          </form>
+          </Button>
+        ) : (
+          <Dropdown
+            isOpen={this.state.dropdownOpen}
+            toggle={this.toggleDropdown}
+          >
+            <DropdownToggle className="text-white btn-block" color="link" size="sm">Change</DropdownToggle>
```
albemarle: Change should be i18n'ed.

src/components/ProfilePage/ProfileAvatar.jsx
abutterworth: We need to match undefined as well I think

src/components/ProfilePage.jsx
davidjoy: This will be fixed on my current branch, yeah.  I cleaned up alllll the various transform/case changing code.

src/components/ProfilePage/ProfileAvatar.jsx
abutterworth: I think this should match undefined as well

src/components/ProfilePage/ProfileAvatar.jsx
abutterworth: 👍

src/components/ProfilePage.jsx
abutterworth: 👍

src/components/ProfilePage/ProfileAvatar.jsx
abutterworth: 👍

src/components/ProfilePage/ProfileAvatar.jsx
albemarle: _.isEmpty() ?

abutterworth: I guess undefined is not a case: going for what you suggested initially ===

PR #51, opened 2019-03-01
Remove editing capability for accounts that requireParentalConsent

abutterworth:


src/selectors/ProfilePageSelector.js
```diff
@@ -29,7 +29,7 @@ export const editableFormModeSelector = createSelector(
     let propExists = account[formId] != null && account[formId].length > 0;
     propExists = formId === 'certificates' ? certificates !== null : propExists; // overwrite for certificates
     // If this isn't the current user's profile...
-    if (!isCurrentUserProfile) {
+    if (!isCurrentUserProfile || account.requiresParentalConsent) {
```
albemarle: The comment on line 31 is now incomplete... but it can probably be deleted.

abutterworth: good call I'll delete it

abutterworth: Actually updated it instead.

PR #53, opened 2019-03-04
Client/server data adapter and simpler data model.

davidjoy:
This PR cleans up our post-API-call data manipulation and simplifies the data structures that we use in the actual reducers/components.  

- Wrote some reusable code to modify object keys in an abstract way.  Made concrete impls for camel case, snake case, and arbitrary replacements.  This can then be used throughout the APIs in a generic way.
- Visibility fields have been flattened out, rather than being under a `visibility: {}` object.
- Moving back towards having ProfilePage.jsx be the sole `connect()`'d component.  It feels a bit wordier, but much simpler/easier to reason.
- Sending account_privacy back to the preferences PATCH correctly.
- Initializing visibilities based on existing account_privacy.
- Fixing profile image bug.
- Fixing social links disappearing bug.
- Fixing empty/null bio/name bug.


config/webpack.dev.config.js
```diff
@@ -25,10 +25,6 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
-          path.resolve(__dirname, 'node_modules/camelcase-keys'),
```
davidjoy: These dependencies have been removed, so no need anymore.

src/actions/ProfileActions.js
```diff
@@ -21,11 +20,11 @@ export const fetchProfileBegin = () => ({
   type: FETCH_PROFILE.BEGIN,
 });
 
-export const fetchProfileSuccess = (account, preferences, certificates) => ({
+export const fetchProfileSuccess = (account, preferences, courseCertificates) => ({
   type: FETCH_PROFILE.SUCCESS,
   account,
   preferences,
-  certificates,
+  courseCertificates,
```
davidjoy: The new agreed-upon name.

src/analytics/analytics.js
```diff
@@ -14,7 +14,7 @@ function handleTrackEvents(eventName, properties) {
 // Sends events to tracking log and downstream
 // TODO: Determine consistent naming for eventName vs eventType and properties v eventData.
 function logEvent(eventType, eventData) {
-  snakecaseKeys(eventData, { deep: true });
+  snakeCaseObject(eventData, { deep: true });
```
davidjoy: Home grown to get rid of problematic dependency.

src/components/ProfilePage.jsx
```diff
@@ -67,17 +67,30 @@ export class ProfilePage extends React.Component {
...
   }
 
-  handleChange(formId, name, value) {
-    if (name === 'visibility') {
-      this.props.updateVisibilityDraft(formId, value);
-    } else {
-      this.props.updateAccountDraft(formId, value);
-    }
+  handleChange(name, value) {
+    this.props.updateDraft(name, value);
```
davidjoy: Only one kind of draft to update now - everything goes to the same place.

src/components/ProfilePage.jsx
```diff
@@ -67,17 +67,30 @@ export class ProfilePage extends React.Component {
...
+      education,
+      visibilityEducation,
+      socialLinks,
+      draftSocialLinksByPlatform,
+      visibilitySocialLinks,
+      languageProficiencies,
+      visibilityLanguageProficiencies,
+      visibilityCourseCertificates,
+      bio,
+      visibilityBio,
```
davidjoy: We pass all the data down to the child forms now, rather than having them select it themselves.  Better symmetry.

src/components/ProfilePage/Bio.jsx
```diff
@@ -27,7 +27,7 @@ class Bio extends React.Component {
 
   handleChange(e) {
     const { name, value } = e.target;
-    this.props.changeHandler(this.props.formId, name, value);
+    this.props.changeHandler(name, value);
```
davidjoy: Since all the names are unique, we don't need to pass formId up here.

src/components/ProfilePage/Bio.jsx
```diff
@@ -61,16 +61,16 @@ class Bio extends React.Component {
...
-                  value={value}
+                  value={bio || ''}
                   invalid={error != null}
                   onChange={this.handleChange}
                 />
                 <FormFeedback>{error}</FormFeedback>
               </FormGroup>
               <FormControls
-                formId={formId}
+                visibilityId="visibilityBio"
```
davidjoy: It's a bit odd that the visibility dropdown is in a class called "FormControls" - it's a bit specific to this profile page's layout.  That said, because it's there, we need to pass in some info for it so it can be given a meaningful ID.

src/components/ProfilePage/SocialLinks.jsx
```diff
@@ -43,24 +43,36 @@ class SocialLinks extends React.Component {
...
-      const updatedList = this.props.committedValue.map((socialLink) => {
-        if (socialLink.platform === name) {
-          return { platform: name, social_link: value };
-        }
-        return socialLink;
-      });
-      this.props.changeHandler(this.props.formId, 'socialLinks', updatedList);
+    const { name, value } = e.target;
+
+    if (name !== 'visibilitySocialLinks') {
```
davidjoy: The social links are a bit special.  If we're updating them, we need to merge them with any existing social link drafts, essentially sending a fresh copy of the whole data structure back to the reducer.  This helps the reducer stay simple and keeps special cases out of it, concentrating them here, where they began.

src/components/ProfilePage/elements/FormControls.jsx
```diff
@@ -22,20 +24,19 @@ function FormControls({
           id={visibilityId}
           className="w-auto"
           type="select"
-          name="visibility"
+          name={visibilityId}
```
davidjoy: Name should be unique now - before we were just sending 'visibility' up and relying on the formId up there to know which one it was.

src/sagas/RootSaga.js
```diff
@@ -66,35 +69,60 @@ export function* handleFetchProfile(action) {
 }
 
 export function* handleSaveProfile(action) {
-  const { username, accountDrafts, visibilityDrafts } = yield select(handleSaveProfileSelector);
-
   try {
+    const { username, drafts, preferences } = yield select(handleSaveProfileSelector);
+
+    const accountDrafts = keepKeys(drafts, [
```
davidjoy: To save the profile, we split our drafts in two so they can be sent to the correct endpoints.

src/sagas/RootSaga.js
```diff
@@ -66,35 +69,60 @@ export function* handleFetchProfile(action) {
...
+      'visibilityCourseCertificates',
+      'visibilityCountry',
+      'visibilityEducation',
+      'visibilityLanguageProficiencies',
+      'visibilityName',
+      'visibilitySocialLinks',
+    ]);
+
+    if (Object.keys(preferencesDrafts).length > 0) {
+      preferencesDrafts.accountPrivacy = 'custom';
```
davidjoy: This is a special little nugget.  If we've updated any visibilities at all, we set account privacy to 'custom'.

src/sagas/RootSaga.js
```diff
@@ -66,35 +69,60 @@ export function* handleFetchProfile(action) {
...
-      visibility: visibilityDrafts,
-    };
 
     if (Object.keys(accountDrafts).length > 0) {
       accountResult = yield call(ProfileApiService.patchProfile, username, accountDrafts);
     }
 
-    if (Object.keys(visibilityDrafts).length > 0) {
-      yield call(ProfileApiService.patchPreferences, username, preferences);
+    let preferencesResult = preferences; // assume it hasn't changed.
```
davidjoy: If we don't make the prefs call (because there were no changes), just assume the "result" is what our preferences currently are.

src/selectors/ProfilePageSelector.js
```diff
@@ -69,26 +72,9 @@ export const formErrorSelector = createSelector(
...
-    formId,
-    error,
-    saveState,
-    accountDraftsField,
-    visibilityDraftsField,
-  ) => ({
-    value: accountDraftsField || account[formId] || '',
-    committedValue: account[formId] || '',
-    visibility: visibilityDraftsField || visibility[formId] || 'private',
+  (editMode, error, saveState) => ({
```
davidjoy: In the process of removing these last three pieces of data for the forms, then we can disconnect them again.  It was well intentioned, but created weird asymmetry in the data flow.

src/selectors/ProfilePageSelector.js
```diff
@@ -97,24 +83,231 @@ export const editableFormSelector = createSelector(
...
+  return draft !== undefined ? draft : committed;
+}
+
+export const formValuesSelector = createSelector(
+  profileAccountSelector,
+  visibilitiesSelector,
+  profileDraftsSelector,
+  profileCourseCertificatesSelector,
+  formSocialLinksSelector,
+  (account, visibilities, drafts, courseCertificates, socialLinks) => ({
```
davidjoy: It's possible this selector could be dried up further, but I didn't want to overdo it.  This way you can see allllll the fields laid out before you.

src/selectors/ProfilePageSelector.js
```diff
@@ -97,24 +83,231 @@ export const editableFormSelector = createSelector(
...
+  draftSocialLinksByPlatformSelector,
+  (
+    account,
+    formValues,
+    profileImage,
+    saveState,
+    savePhotoState,
+    isCurrentUserProfile,
+    draftSocialLinksByPlatform,
+  ) => ({
```
davidjoy: Similar to the above comment here - could be DRY'd up further with some `...` operators, but seeing all the data and knowing exactly what's being given to the component has its own merits.

src/components/ProfilePage.jsx
abutterworth: party!

src/components/ProfilePage/Bio.jsx
abutterworth: would you prefer passing in the whole <VisibilitySelect /> in as a prop to FormControls?
`visibilitySelect={<VisibilitySelect id="visibilityBio" />}`

src/components/ProfilePage.jsx
```diff
@@ -112,11 +125,37 @@ export class ProfilePage extends React.Component {
...
+              />
+              <Education
+                education={education}
+                visibilityEducation={visibilityEducation}
+                formId="education"
+                {...commonFormProps}
+              />
+              <SocialLinks
+                socialLinks={socialLinks}
+                draftSocialLinksByPlatform={draftSocialLinksByPlatform}
```
abutterworth: The other forms get their drafts straight from the store. Is there a reason to pass socialLinks in from the parent?

abutterworth: Oh nvm, the others are different than I first thought.

davidjoy: They all have their values passed in from ProfilePage.jsx.  The selector for ProfilePage.jsx determines if it should use the draft value (if it exists) or the committed value.  SocialLinks gets it's drafts/committed values passed in on line 153 there - `socialLinks={socialLinks}`.

In addition to that, SocialLinks.jsx needs something extra, which is this drafts lookup.  When editing social links, it wants to push a complete drafts object back to the reducer _in the same order_ as before.  `draftSocialLinksByPlatform` is just a lookup of the links by their platform name so that we can easily reconstruct the drafts data onChange in the same order it was originally in - you can see where it's used in the file itself.

src/selectors/ProfilePageSelector.js
```diff
@@ -97,24 +83,231 @@ export const editableFormSelector = createSelector(
...
+        };
+    }
+  },
+);
+
+/**
+ * If there's no draft present at all (undefined), use the original committed value.
+ */
+function chooseFormValue(draft, committed) {
+  return draft !== undefined ? draft : committed;
```
abutterworth: `typeof draft !== 'undefined'`?

src/components/ProfilePage/Bio.jsx
davidjoy: I think it would make more sense to pull it out completely.  It renders on a row above the buttons - there's not much reason it needs to be in there, I'd think.  Using Bio as an example, it's a form element that's effectively a peer of the textarea input... having it bundled with the form controls themselves seems odd, you know?

src/selectors/ProfilePageSelector.js
davidjoy: Not really necessary to my knowledge... if the value of draft is `undefined` then its type will be `'undefined`' they're essentially the same, but this one's a bit shorter. :)

PR #54, opened 2019-03-04
Better error handling

abutterworth:
- Save buttons never appear in a "save failed" state
- errors appear in the UI 
- Social Links errors appear at the top of the set of links (API needs update if we want per-field errors)

### DIFF
--- a/src/sagas/RootSaga.js
+++ b/src/sagas/RootSaga.js
@@ -127,8 +127,8 @@ export function* handleSaveProfile(action) {
     yield delay(300);
     yield put(saveProfileReset());
     yield put(resetDrafts());
-  } catch ({ fieldErrors }) {
-    yield put(saveProfileFailure(fieldErrors));
+  } catch (e) {
+    yield put(saveProfileFailure(e.fieldErrors));
   }
 }
 
--- a/src/sagas/RootSaga.js
+++ b/src/sagas/RootSaga.js
@@ -127,8 +127,8 @@ export function* handleSaveProfile(action) {
     yield delay(300);
     yield put(saveProfileReset());
     yield put(resetDrafts());
-  } catch (e) {
-    yield put(saveProfileFailure(e.message));
+  } catch ({ fieldErrors }) {
+    yield put(saveProfileFailure(fieldErrors));
   }
 }
 
--- a/src/services/ProfileApiService.js
+++ b/src/services/ProfileApiService.js
@@ -37,7 +37,16 @@ export async function patchProfile(username, params) {
         'Content-Type': 'application/merge-patch+json',
       },
     },
-  );
+  ).catch((error) => {
+    const processedError = Object.create(error);
+    const fieldErrors = Object.entries(processAccountData(error.response.data.field_errors))
+      .reduce((acc, [fieldKey, messages]) => {
+        acc[fieldKey] = messages.userMessage;
+        return acc;
+      }, {});
+    processedError.fieldErrors = fieldErrors;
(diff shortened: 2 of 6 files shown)


src/services/ProfileApiService.js
```diff
@@ -37,7 +37,16 @@ export async function patchProfile(username, params) {
...
     },
-  );
+  ).catch((error) => {
+    const processedError = Object.create(error);
+    const fieldErrors = Object.entries(processAccountData(error.response.data.field_errors))
+      .reduce((acc, [fieldKey, messages]) => {
+        acc[fieldKey] = messages.userMessage;
+        return acc;
+      }, {});
+    processedError.fieldErrors = fieldErrors;
```
davidjoy: Might just call this "errors" since that's what it's becoming a step or two later anyway... that way the name just stays consistent.  And it's shorter that way. ;)

src/services/ProfileApiService.js
```diff
@@ -37,7 +37,16 @@ export async function patchProfile(username, params) {
         'Content-Type': 'application/merge-patch+json',
       },
     },
-  );
+  ).catch((error) => {
```
davidjoy: So we should find a way to make this catch generic, probably, so it's reusable amongst all our API endpoints.  (And then hold the backend code to a standard error response!)

src/sagas/RootSaga.js
```diff
@@ -127,8 +127,8 @@ export function* handleSaveProfile(action) {
     yield delay(300);
     yield put(saveProfileReset());
     yield put(resetDrafts());
-  } catch (e) {
-    yield put(saveProfileFailure(e.message));
+  } catch ({ fieldErrors }) {
```
davidjoy: Should we just pass the whole error object along here?  Or include the non-field errors as well?  (I'm assuming that we might receive some non-field-specific errors too, sometimes...)

src/services/ProfileApiService.js
```diff
@@ -37,7 +37,16 @@ export async function patchProfile(username, params) {
...
+    const processedError = Object.create(error);
+    const fieldErrors = Object.entries(processAccountData(error.response.data.field_errors))
```
davidjoy: Is it possible for there to be more than one error per field?

davidjoy: (Since it was an array, it seems like it's at least plausible...)

src/sagas/RootSaga.js
abutterworth: If we add non-field error stuff I think we should add it to the object deconstruction. I wanted to be explicit about what we're working with. In the end, either way works.

src/services/ProfileApiService.js
abutterworth: It doesn't appear that we get arrays back. I tried by making two errors in social links and I only get the first

src/services/ProfileApiService.js
abutterworth: 👍

src/services/ProfileApiService.js
abutterworth: fieldErrors is clearer, no? I actually think maybe field errors is a better name throughout.

robrap:
@abutterworth @davidjoy: FYI: You will soon be able to log errors or info to newrelic as well without interrupting the user flow.  I didn't look to see if that is or is not appropriate here, but you can decide.

abutterworth:
Merging without changing the name of fieldErrors. If you feel strongly, I'm not concerned if you modify.

PR #55, opened 2019-03-04
Certificate Illustration Backgrounds

abutterworth:
Default is verified certificate illustration.


src/index.scss
```diff
@@ -114,10 +114,26 @@ $fa-font-path: "~font-awesome/fonts";
...
       font-family: $font-family-serif;
       font-weight: 400;
     }
+    .certificate-type-illustration {
+      position: absolute;
+      top: 0;
+      left: 50%;
+      right: 1rem;
+      bottom: 0;
+      // z-index: -1;
```
albemarle: Should this be deleted?

abutterworth: Yes good catch

PR #56, opened 2019-03-05
ARCH-430: Add logEvent error handling and NewRelic.

robrap:
Add Logging Service and event fixes.
    
    - Original NewRelicService lifted from edx-portal.
    - Fixed bug with snake case of event data.
    - Added error handling for logEvent.
    - Added tests for logEvent.
    
[ARCH-430](https://openedx.atlassian.net/browse/ARCH-430)

FYI: This includes a temporary rename of the event so it can go to Production.  If I need to move that commit to another branch, I can do that.  Otherwise, I'll just land this whole thing together.

FYI: @edx/arch-squad


src/services/LoggingService.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  static logError(error) {
+    if (typeof newrelic !== 'undefined') {
+      newrelic.noticeError(error);
+    }
+  }
+
+  static logAPIErrorResponse(error) {
+    let { message } = error;
```
albemarle: Can also do ```const {message, response, request} = error;``` so you don't have to repeat so much.

robrap: Thanks.  Code copied from edx-portal.  Going to merge to unblock prod.  I like the idea though.

PR #63, opened 2019-03-07
Style 404 Page

abutterworth:
- Center text
- Uses inline styles


albemarle:
I believe the style() semantic commit category is for changes that only change how your code is formatted: https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

PR #64, opened 2019-03-07
Do not show age message when viewing other's profiles

abutterworth:
- Also, remove the close button on the message

### DIFF
--- a/src/components/ProfilePage.jsx
+++ b/src/components/ProfilePage.jsx
@@ -92,6 +92,8 @@ export class ProfilePage extends React.Component {
       visibilityCourseCertificates,
       bio,
       visibilityBio,
+      requiresParentalConsent,
+      isCurrentUserProfile,
     } = this.props;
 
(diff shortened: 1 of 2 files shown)


abutterworth:
It's worth noting here, that the client does receive a boolean for `requiresParentalConsent` when requesting other's profiles. This boolean is true when the age is not set or if the age is under 13. It's not really a privacy risk, but we should probably not return this information at all.

src/components/ProfilePage.jsx
```diff
@@ -101,6 +103,8 @@ export class ProfilePage extends React.Component {
       changeHandler: this.handleChange,
     };
 
+    const shouldShowAgeMessage = requiresParentalConsent && isCurrentUserProfile;
```
albemarle: Nice, I like this.

abutterworth: I was inspired by your earlier feedback

PR #66, opened 2019-03-11
fix(a11y): Improvements to form descriptions and heading levels

abutterworth:


src/components/ProfilePage/Bio.jsx
```diff
@@ -54,27 +54,32 @@ class Bio extends React.Component {
...
-              <FormGroup>
-                <Label for={formId}>{intl.formatMessage(messages['profile.bio.about.me'])}</Label>
-                <Input
-                  type="textarea"
-                  id={formId}
-                  name={formId}
-                  value={bio || ''}
-                  invalid={error != null}
-                  onChange={this.handleChange}
+            <div role="dialog" aria-labelledby={`${formId}-label`}>
```
albemarle: The British spelling of "labelled" always makes me do a double take.  Better than "Referer", I guess.  :-P

abutterworth: Yea, I only know it's correct if it's the right colour in the IDE ;)

albemarle: I see what you did there!

PR #71, opened 2019-03-11
fix(certificates): handle absolute and relative certificate download urls

douglashall:
The certificates API can return both absolute and relative download URLs depending on whether or not the certificate is PDF or web.


douglashall:
@abutterworth Good point. I will update. Thanks!

albemarle:
Not a merge blocker, but this would be a nice thing to have a test for.

PR #73, opened 2019-03-11
ARCH-390: Fix analytics events to be x-www-form-urlencoded.

robrap:
The LMS /event api expects the request to use a content type of
application/x-www-form-urlencoded.

ARCH-390

FYI: @edx/arch-squad

### DIFF
--- a/src/analytics/analytics.js
+++ b/src/analytics/analytics.js
@@ -5,6 +5,19 @@ import LoggingService from '../services/LoggingService';
 
 const eventLogApiBaseUrl = `${configuration.LMS_BASE_URL}/event`;
 
+
+// Uses a JSON as a source of key value pairs for x-www-form-urlencoded output.
+// Note should be handled by axios, but is not built in.
+// - see https://github.com/axios/axios/issues/97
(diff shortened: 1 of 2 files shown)


src/analytics/analytics.js
```diff
@@ -5,6 +5,19 @@ import LoggingService from '../services/LoggingService';
 
 const eventLogApiBaseUrl = `${configuration.LMS_BASE_URL}/event`;
 
+
+// Uses a JSON as a source of key value pairs for x-www-form-urlencoded output.
+// Note should be handled by axios, but is not built in.
+// - see https://github.com/axios/axios/issues/97
```
robrap: @douglashall: Not sure if we can move this into frontend-auth apiClient setup?  I'll discuss more.

douglashall: I'd probably go with something like this instead: https://www.npmjs.com/package/form-urlencoded

robrap: @douglashall: I’ll discuss tomorrow the possible use of this library for a more general solution with axios. Merged this as-is to fix big and can iterate. Thanks.

PR #77, opened 2019-03-11
i18n: add make targets for Transifex

albemarle:


abutterworth:
Is there a reason to do this workin in a Makefile instead of npm scripts?

albemarle:
The existing pipelines that this plugs into use make.

PR #78, opened 2019-03-12
ARCH-390: Use form-urlencoded npm package.

robrap:
Switch from custom function to form-urlencoded to perform form url
encoding.

ARCH-390

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9208,6 +9208,11 @@
         "mime-types": "^2.1.12"
       }
     },
+    "form-urlencoded": {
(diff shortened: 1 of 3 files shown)


robrap:
@douglashall: FYI:
1. The test still passes, so should be fine.
2. Bootstrap commit hash changed in my lock file, so I undid that change.  Not sure why that happens.
3. The rest of the lock updates should be fine, but not sure.  My npm version is `6.5.0`.  Not sure if we are attempting to standardize?

PR #83, opened 2019-03-13
fix: remove the upload button background when viewing other profiles

abutterworth:
Also removes flipping the isLoadingProfile state when fetching a profile begins.

Desired behavior: Any time a whole profile is being loaded show the loading state. 

Right now the upload photo endpoint doesn't return the url we need so we refetch the whole profile again. This causes the whole page to appear to reload.

For now removing the resetting of isLoadingProfile to true inside FETCH_PROFILE.BEGIN will mean that isLoadingProfile is only true when the app first loads (it is the default state). We can uncomment this line in the future when we update the upload profile endpoint.

### DIFF
--- a/src/components/ProfilePage/ProfileAvatar.jsx
+++ b/src/components/ProfilePage/ProfileAvatar.jsx
@@ -60,52 +60,60 @@ class ProfileAvatar extends React.Component {
   renderMenu() {
     if (!this.props.isEditable) return null;
 
-    if (this.props.isDefault) {
-      return (
-        <Button
-          className="text-white btn-block"
-          color="link"
-          size="sm"
-          onClick={this.onClickUpload}
-        >
-          <FormattedMessage
-            id="profile.profileavatar.upload-button"
-            defaultMessage="Upload Photo"
-            description="Upload photo button"
-          />
-        </Button>
-      );
-    }
-
-    return (
-      <Dropdown
-        isOpen={this.state.dropdownOpen}
-        toggle={this.toggleDropdown}
-      >
-        <DropdownToggle className="text-white btn-block" color="link" size="sm">
-          <FormattedMessage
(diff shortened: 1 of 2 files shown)


src/reducers/ProfilePageReducer.js
```diff
@@ -28,7 +28,10 @@ const profilePage = (state = initialState, action) => {
     case FETCH_PROFILE.BEGIN:
       return {
         ...state,
-        isLoadingProfile: true,
+        // TODO: uncomment this line after ARCH-438 Image Post API returns the url
```
albemarle: Nice TODO!

src/components/ProfilePage/ProfileAvatar.jsx
```diff
@@ -60,62 +60,68 @@ class ProfileAvatar extends React.Component {
...
-        <DropdownToggle className="text-white btn-block" color="link" size="sm">
-          <FormattedMessage
-            id="profile.profileavatar.change-button"
-            defaultMessage="Change"
-            description="Change photo button"
-          />
-        </DropdownToggle>
-        <DropdownMenu>
-          <DropdownItem onClick={this.onClickUpload}>
+    const menuContent = ((isDefault) => {
```
davidjoy: This pattern of creating anonymous functions and passing the props in feels unnecessarily complex to me... It's not unreasonable to just use `this.props.isDefault` directly, and/or to just write these as instance methods.  It feels a bit odd that we have two different ways of writing render methods - one using instance methods and one with these anonymous functions.  

While technically in Javascript the instance methods are not 'private', practically speaking no one outside the component will ever call them nor should, so it feels okay to me to just create helper instance methods.  It keeps the code a bit flatter/keeps each method succinct.

davidjoy:
😻

PR #86, opened 2019-03-13
Update page title to match existing profile.

davidjoy:
Learner Profile | edX

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,9 +1,9 @@
 <!doctype html>
 <html lang="en-us">
   <head>
-      <title>Account | edX</title>
+      <title>Learner Profile | edX</title>
       <meta charset="utf-8">
-      <meta name="viewport" content="width=device-width, initial-scale=1.0">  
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -1,9 +1,9 @@
 <!doctype html>
 <html lang="en-us">
   <head>
-      <title>Account | edX</title>
+      <title>Learner Profile | edX</title>
```
albemarle: Should this be i18n'ed?

davidjoy: :O  That's a very good point.  We could dynamically update it once we get the translations...

albemarle: If we do this, the string should include the | edX part, so that the translators can flip it for RTL.

abutterworth: If we're going to dynamically update it should we put the user's name or username in there?

davidjoy: @abutterworth Possibly - it doesn't do that currently, fwiw.

PR #89, opened 2019-03-13
fix: deal with Transifex id escaping

albemarle:
Also renamed some variables to try to make the code clearer.


src/i18n/i18n-concat.js
```diff
@@ -17,28 +21,38 @@ function gatherJson(dir) {
   return ret;
 }
 
+// the hash file returns ids whose periods are "escaped" (sort of), like this:
+// "key": "profile\\.sociallinks\\.social\\.links"
```
abutterworth: strange!

abutterworth:
An aside: We may want to use something to parse command line arguments to make this less brittle in the future. (something like https://github.com/tj/commander.js/)
`if (process.argv[3] === '--comments') { // prepare to handle the translator notes`
https://github.com/edx/frontend-app-profile/blob/e0763935b83b384df1106c3042bd61708948d3c6/src/i18n/i18n-concat.js#L34

albemarle:
@abutterworth Yeah there are many improvements that can be made, for sure.  (Including renaming that script!) It's based off the reactifex script originally, though at this point is almost unrecognizable.

PR #93, opened 2019-03-14
ARCH-427: Fix and log more unhandled errors.

robrap:
- Logs errors that fall through to the App.
- Logs unprocessed error in Saga.
- Ensures ProfileApiService throws proper errors.
- Fix LoggingService to pass through non-API errors.
- Fix LoggingService to cap length of message.

ARCH-427

FYI: @edx/arch-squad

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -12,7 +12,6 @@ import { handleTrackEvents } from '../analytics/analytics';
 import { getLocale, getMessages } from '../i18n/i18n-loader';
 import SiteHeader from './common/SiteHeader';
 import ConnectedProfilePage from './ProfilePage';
-import LoggingService from '../services/LoggingService';
 
 import FooterLogo from '../../assets/edx-footer.png';
 import NotFoundPage from './NotFoundPage';
@@ -22,16 +21,6 @@ class App extends Component {
     const { username } = this.props;
     const userAccountApiService = new UserAccountApiService(apiClient, process.env.LMS_BASE_URL);
     this.props.fetchUserAccount(userAccountApiService, username);
-  }
-
-  componentDidCatch(error, info) {
-    let processedError = error;
-    if (info) {
-      processedError = Object.create(error);
-      processedError.message = `${error.message} ${JSON.stringify(info)}`;
-    }
-    LoggingService.logAPIErrorResponse(processedError);
-    // TODO: Handle user UX for uncaught errors
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


src/services/LoggingService.js
```diff
@@ -12,20 +12,24 @@ class LoggingService {
...
   static logAPIErrorResponse(error) {
-    let { message } = error;
+    let processedError = Object.create(error);
     if (error.response) {
-      message = `${error.response.status} ${error.response.config.url} ${JSON.stringify(error.response.data)}`;
+      processedError = new Error(`API request failed: ${error.response.status} ${error.response.config.url} ${JSON.stringify(error.response.data)}`);
     } else if (error.request) {
-      message = `${error.request.status} ${error.request.responseURL} ${error.request.responseText}`;
-    } else if (error.stack) {
-      message = error.stack;
```
robrap: FYI: Errors with stack are now passed through, instead of creating a new Error, so I _think_ this should be ok.

src/services/LoggingService.js
```diff
@@ -12,20 +12,24 @@ class LoggingService {
...
+      processedError = new Error(`API request failed: ${error.request.status} ${error.request.responseURL} ${error.request.responseText}`);
     }
-    this.logError(new Error(`API request failed: ${message}`));
+    if (processedError.message) {
+      // NewRelic will not log the error if it is too long.
+      processedError.message = processedError.message.substring(0, 4000);
```
robrap: FYI: We had places where `error.response.data` was an entire HTML 500 page, and the message got so long that nothing was sent to NewRelic and the whole things was lost.  This limit fixes that issue.
FYI: @georgebabey

georgebabey: Good catch - thanks @robrap

src/services/LoggingService.js
georgebabey: Should LoggingService be pulled out into a separate library so we can all share this code rather than having different copies with different implementations?

robrap: @georgebabey: As we pull this and some other utilities into the Cookie Cutter, we'll have to decide whether to create a package or not.  There are some trade-offs either way, but I lean toward packages.  Maybe there is a decision tree we could slowly put together to help us decide?  This is a good fedx topic both generally, as well as more specifically regarding logging, analytics, i18n, other?

src/components/App.jsx
```diff
@@ -23,6 +24,16 @@ class App extends Component {
     this.props.fetchUserAccount(userAccountApiService, username);
   }
 
+  componentDidCatch(error, info) {
```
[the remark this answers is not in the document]
abutterworth: As discussed this is a React method: https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html

src/services/LoggingService.js
albemarle: Would be nice to make that a well-named constant.

robrap: @albemarle: Agreed and good point.  Will add that to the refactor.  Thanks.

PR #95, opened 2019-03-14
Shut off name editing and tweak visibility spacing

davidjoy:
As per @nasthagiri, shutting off name editing.

Also fixing a layout ugliness I noticed in the visibility.  The eyeball icon was a little too cozy with the text label.


davidjoy:
Note that this PR leaves the plumbing for name editing in.  I want to shut it off for now and then follow up and fix that once we're 100% that it won't be coming back in one form or another.

albemarle:
I would probably favor deleting it and then using git to get it back if we needed to do that, but okay either way.

PR #102, opened 2019-03-15
Call identifyUser before sending page events.

douglashall:


src/index.jsx
```diff
@@ -16,7 +16,6 @@ if (apiClient.ensurePublicOrAuthencationAndCookies(window.location.pathname)) {
 
   ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
 
-  // identify user for future analytics calls
-  // TODO: Call before each page call.
   identifyUser();
+  window.analytics.page();
```
robrap: Heh.  Simple enough.  I think the idea was to hide Segment inside analytics.  So, you should just add a light wrapper (like `identifyUser`) and call it here.  Maybe named `pageEvent` (unless you have a better idea)?

src/analytics/analytics.js
```diff
@@ -46,5 +45,9 @@ function identifyUser() {
   }
 }
 
+function sendPageEvent() {
```
robrap: Nice.  And smart.  I'd like to add a little documentation, but maybe when this moves to frontend-analytics.  Thanks.

FYI: I also confirmed with Data Science that for pages that require authentication, we do **not** need a page event here for the unauthenticated user, and should just log events for what the user sees, which would be an unauthenticated view of the login page with the profile page as the referrer.

PR #103, opened 2019-03-15
ARCH-578: Enhance the logging service.

robrap:
- Add custom attributes to api response errors.
- Add custom attribute capabilities to loadError.
- Drop the data for html page responses.
- Move code for max message length.
- Add tests.

ARCH-578

FYI: @edx/arch-squad

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "fetch-mock": "^6.3.0",
     "file-loader": "^1.1.9",
     "html-webpack-harddisk-plugin": "^0.2.0",
-    "html-webpack-new-relic-plugin": "^1.0.1",
+    "html-webpack-new-relic-plugin": "git+https://git@github.com/edx/html-webpack-new-relic-plugin.git#e7204801b9484faa38b448bef3e6a8a3bfa05b6b",
     "html-webpack-plugin": "^3.0.3",
     "husky": "^0.14.3",
     "identity-obj-proxy": "^3.0.0",
--- a/src/services/LoggingService.js
+++ b/src/services/LoggingService.js
@@ -3,6 +3,21 @@
  *
  * Requires the NewRelic Browser JavaScript snippet.
  */
+
+// NewRelic will not log an error if it is too long.
+const MAX_ERROR_LENGTH = 4000;
+
+function fixErrorLength(error) {
+  if (error.message && error.message.length > MAX_ERROR_LENGTH) {
+    const processedError = Object.create(error);
+    processedError.message = processedError.message.substring(0, MAX_ERROR_LENGTH);
+    return processedError;
+  } else if (typeof error === 'string' && error.length > MAX_ERROR_LENGTH) {
+    return error.substring(0, MAX_ERROR_LENGTH);
+  }
+  return error;
+}
+
 class LoggingService {
   static logInfo(message) {
     if (typeof newrelic !== 'undefined') {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -90,7 +90,7 @@
     "fetch-mock": "^6.3.0",
     "file-loader": "^1.1.9",
     "html-webpack-harddisk-plugin": "^0.2.0",
-    "html-webpack-new-relic-plugin": "^1.0.1",
+    "html-webpack-new-relic-plugin": "git+https://git@github.com/edx/html-webpack-new-relic-plugin.git#e7204801b9484faa38b448bef3e6a8a3bfa05b6b",
```
robrap: @douglashall: FYI: This is what is needed until we either get the fix merged upstream, or just create our own repo with our own package.

src/services/LoggingService.js
```diff
@@ -3,36 +3,67 @@
...
-      processedError.message = processedError.message.substring(0, 4000);
-    }
-    this.logError(processedError);
+
+    this.logError(processedError, updatedCustomAttributes);
   }
 }
 
 export default LoggingService;
+export { MAX_ERROR_LENGTH };
```
robrap: Reviewers: Is it wrong to export only for a test?  Should I just duplicate the constant?

src/services/LoggingService.js
```diff
@@ -3,36 +3,67 @@
...
       newrelic.addPageAction('INFO', { message });
     }
   }
 
-  static logError(error) {
+  static logError(error, customAttributes) {
     if (typeof newrelic !== 'undefined') {
       // Note: customProperties are not sent.  Presumably High-Security Mode is being used.
-      newrelic.noticeError(error);
+      newrelic.noticeError(fixErrorLength(error), customAttributes);
```
robrap: @abutterworth: FYI: The problem we were having with custom attributes was that it is a newer feature and wasn't supported by the version of the newrelic snippet we had installed.  The npm package update brings in a newer snippet that supports these custom attributes.

src/services/LoggingService.js
```diff
@@ -3,36 +3,67 @@
...
+      // Don't include data if it is just an HTML document, like a 500 error page.
+      data = data.includes('<!DOCTYPE html>') ? '' : data;
+      const responseAttributes = {
+        errorType: 'api-response-error',
+        errorStatus: error.response.status,
+        errorUrl: error.response.config ? error.response.config.url : '',
+        errorData: data,
+      };
+      updatedCustomAttributes = Object.assign({}, customAttributes, responseAttributes);
+      processedError = new Error(`API request failed: ${error.response.status} ${responseAttributes.errorUrl} ${data}`);
```
robrap: @abutterworth: FYI: I added status code but in the NewRelic GUI it takes too much room and doesn't allow you to see the differences between the errors, so I opted to drop it.

src/services/LoggingService.js
```diff
@@ -3,36 +3,67 @@
  *
  * Requires the NewRelic Browser JavaScript snippet.
  */
+
+// NewRelic will not log an error if it is too long.
+const MAX_ERROR_LENGTH = 4000;
```
robrap: @abutterworth @georgebabey: FYI: I left this in because I think it is better to have some of the message than none of the message.  However, I did learn that the event is logged either way (with the newer newrelic snippet and possibly with the older one), but the default view is filtered on message and the message itself is not sent so the error is filtered from that view.  If you switch to `errorClass`, then you might be able to see the additional errors.

src/services/LoggingService.js
davidjoy: If we're being super-purists about it then we'd say it's wrong. :)  Practically speaking, no one will ever really notice.  If if were in a separate constants file and imported here, then it wouldn't feel as wrong, but I wouldn't bother.

src/services/LoggingService.js
```diff
@@ -3,36 +3,67 @@
...
+      let data = !error.response.data ? '' : JSON.stringify(error.response.data);
+      // Don't include data if it is just an HTML document, like a 500 error page.
+      data = data.includes('<!DOCTYPE html>') ? '' : data;
+      const responseAttributes = {
+        errorType: 'api-response-error',
+        errorStatus: error.response.status,
+        errorUrl: error.response.config ? error.response.config.url : '',
+        errorData: data,
+      };
+      updatedCustomAttributes = Object.assign({}, customAttributes, responseAttributes);
```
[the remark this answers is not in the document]
robrap: Swapped.  Thanks.

PR #108, opened 2019-03-19
fix: use lodash.get to provide default message ids to intl.formatMessage

abutterworth:
intl.formatMessage will error out if it's provided a message id that doesn't exist. This PR adds default ids to certificate and education level strings.

### DIFF
--- a/src/components/ProfilePage/Certificates.jsx
+++ b/src/components/ProfilePage/Certificates.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
 import { Row, Col, Card, CardBody, CardTitle, Button, Form } from 'reactstrap';
 import { connect } from 'react-redux';
+import get from 'lodash.get';
 
 import messages from './Certificates.messages';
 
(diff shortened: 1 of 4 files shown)


src/components/ProfilePage/Certificates.jsx
```diff
@@ -74,7 +75,11 @@ class Certificates extends React.Component {
           <CardBody className="d-flex flex-column">
             <CardTitle>
               <p className="small mb-0">
-                {intl.formatMessage(messages[`profile.certificates.types.${certificateType}`])}
+                {intl.formatMessage(get(
+                  messages,
+                  `profile.certificates.types.${certificateType}`,
+                  messages['profile.certificates.types.verified'],
```
douglashall: I'm not sure it is appropriate to default to verified here. Do we know what the certificate type was that was triggering the error?

abutterworth: No I didn't check :/ – but we need a default anyway. Maybe we just add one that says 'Certificate'? Or nothing?

robrap: Agreed.
1. We should fix the actual problem by adding a certificate type for whatever caused the error.
2. Tagging @marcotuts in case he wants to weigh-in on the default in case this shows up in the future.

douglashall: It looks like some of us have certificates with a type of "audit" on stage which I think is what was causing the issue. @abutterworth Yeah, let's create a separate default certificate type translation id for unknown types. "Certificate" makes sense to me as the default translation string.

PR #113, opened 2019-03-22
Display certificate issue date on certificate cards.

douglashall:
Depends on https://github.com/edx/edx-platform/pull/20045.

See it in action at: https://profile.edxdev.org/u/verified

### DIFF
--- a/src/components/ProfilePage/Certificates.jsx
+++ b/src/components/ProfilePage/Certificates.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
+import { injectIntl, intlShape, FormattedDate, FormattedMessage } from 'react-intl';
 import { Row, Col, Card, CardBody, CardTitle, Button, Form } from 'reactstrap';
 import { connect } from 'react-redux';
 import get from 'lodash.get';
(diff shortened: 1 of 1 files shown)


src/components/ProfilePage/Certificates.jsx
```diff
@@ -91,6 +91,14 @@ class Certificates extends React.Component {
             </p>
             <p className="h6 mb-4">{courseOrganization}</p>
             <div className="flex-grow-1" />
+            <p className="small mb-2">
+              <FormattedMessage
+                id="profile.certificate.completion.date.label"
+                defaultMessage="Completed on"
```
robrap: I think you want something like: `"Completed on {date}"`.  See [this example from footer](https://github.com/edx/frontend-component-footer/blob/66f6832887f3cf34b1a32e929a1adfda862c4a49/src/lib/components/SiteFooter/index.jsx#L276-L281).

abutterworth: It should probably also use FormattedDate. Maybe something like this?
```
<FormattedMessage
  id="profile.certificate.completion.date.label"
  defaultMessage="Completed on {date}"
  values={{
    date: <FormattedDate value={new Date(modifiedDate)}/>,
  }}
/>
```
FormattedDate Reference: https://github.com/yahoo/react-intl/wiki/Components#formatteddate

douglashall: Thanks! Pushing update now.

PR #118, opened 2019-03-27
ARCH-566 Display error page when unexpected errors occur.

douglashall:
See it in action here by attempting to change profile data: https://profile.edxdev.org/u/verified

### DIFF
--- a/src/components/ErrorPage.jsx
+++ b/src/components/ErrorPage.jsx
@@ -1,0 +1,46 @@
+import React, { Component } from 'react';
+import { Button, Col, Container, Row } from 'reactstrap';
+import { FormattedMessage } from 'react-intl';
+import { Link } from 'react-router-dom';
+
+import apiClient from '../config/apiClient';
+
+export default class ErrorPage extends Component {
+  componentDidMount() {}
+
+  render() {
+    const { username } = apiClient.getAuthenticationState().authentication;
+
+    return (
+      <Container
+        fluid
+        className="py-5 justify-content-center align-items-start text-center"
+      >
+        <Row>
+          <Col>
+            <p className="my-0 py-5 text-muted">
+              <FormattedMessage
+                id="profile.error.message.text"
+                defaultMessage="An unexpected error occurred. Please click the button below to return to your profile and try again."
+                description="error message when an unexpected error occurs"
+              />
+            </p>
+          </Col>
+        </Row>
+        <Row>
+          <Col>
+            <Link to={`/u/${username}`}>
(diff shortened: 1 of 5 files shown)


src/components/ErrorPage.jsx
```diff
@@ -0,0 +1,36 @@
...
+        fluid
+        className="d-flex py-5 justify-content-center align-items-start text-center"
+      >
+        <p
+          className="my-0 py-5 text-muted"
+          style={{ maxWidth: '32em' }}
+        >
+          <FormattedMessage
+            id="profile.error.message"
+            defaultMessage="An unexpected error occurred. Please click {here} to try again."
```
robrap: Nit: I wonder if `{profileLink}` might be more clear, and might affect translation of the rest of the sentence(?) in some languages.

src/sagas/RootSaga.js
```diff
@@ -70,7 +68,7 @@ export function* handleFetchProfile(action) {
     if (e.response.status === 404) {
       yield put(push('/notfound'));
     } else {
-      yield put(fetchProfileFailure(e.message));
+      yield put(push('/error'));
```
robrap: 1. Do we want `LoggingService.logAPIErrorResponse(e)` here?
2. Should `fetchProfileFailure` still exist if it isn't used?

douglashall: 1. We do have it above ^^.
2. No, I will remove it.

src/components/ErrorPage.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { Link } from 'react-router-dom';
+
+import apiClient from '../config/apiClient';
+
+export default class ErrorPage extends Component {
+  componentDidMount() {}
+
+  renderProfileLink(linkText) {
+    const { username } = apiClient.getAuthenticationState().authentication;
+    return <Link to={`/u/${username}`}>{linkText}</Link>;
```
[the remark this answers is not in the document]
douglashall: I don't see an a11y-related things added by HyperLink in the paragon example. Also, I don't think HyperLink works with react-router. I think we are good with what we have here.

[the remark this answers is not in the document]
robrap: Sorry - I just meant something like setting `aria-label`, since the text `here` doesn't really help.  And I see now that Hyperlink is only setting this for external links.

douglashall:
@robrap There was no UX design for this scenario. We can always go back and change this if someone doesn't like it, but this approach will get the job done for now.

PR #120, opened 2019-04-02
feat(test): add snapshot tests for profile page component

abutterworth:
Fixes ARCH-601.

Adds snapshots for the profile page in different states:
- App loading
- viewing own profile
- Saving edited bio
- viewing someone else's profile

Basically I copied the redux store from the app after navigating to a particular state (editing someting for example) and used it as mock data. I can add README documentation on how this is done if we think it will be helpful in the future.

Screen Shot 2019-04-02 at 1 59 37 PM


abutterworth:
Fixed the missing newlines.

As discussed in person, leaving the snapshot on the `<ProfilePage />` component. Pulling the snapshot tests up to the `<App />` level would require more work since that would required getting the connected router working in the jest environment.

PR #122, opened 2019-04-02
ARCH-560 Add RTL capabilities.

douglashall:
See it in action here: https://profile.edxdev.org/u/verified

### DIFF
--- a/src/i18n/i18n-loader.js
+++ b/src/i18n/i18n-loader.js
@@ -41,4 +41,13 @@ const messages = { // current fallback strategy is to use the first two letters 
 
 const getMessages = (locale = getLocale()) => messages[locale];
 
-export { getLocale, getMessages };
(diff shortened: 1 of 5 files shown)


src/i18n/i18n-loader.js
```diff
@@ -41,4 +41,13 @@ const messages = { // current fallback strategy is to use the first two letters
 
 const getMessages = (locale = getLocale()) => messages[locale];
 
-export { getLocale, getMessages };
+const handleRtl = () => {
+  if (getLocale() === 'ar') {
```
abutterworth: Maybe this should be a check on an array of other rtl languages

douglashall: It definitely should once we support more than one language with an RTL dialect. Trying to keep things simple for now.

PR #123, opened 2019-04-03
Use raw html form elements

abutterworth:
This PR shows what it would look like to use raw html for form elements instead of Reactstrap. Using this as an experiment to determine if we should bring Reactstrap form components into Paragon.

My take: I think we should not include Reactstrap. HTML alleviates the need for another dependency, keeps developers closer to the html (I think a good thing), and gets the job done despite some extra bloat from bootstrap class names. Paragon should add documentation for these class names to prevent developers from bouncing back and forth between bootstrap and paragon docs.

Instead of building input, textarea, select, etc as React components I think we should rely on linters for a11y enforcement and React container components like a [Form Group (example below)](https://github.com/edx/frontend-app-profile/compare/abutterworth/use-raw-html-form-elements?expand=1#diff-dc36925cd4e3904a94498a7b2d93bf0dR63) to handle the construction of label + form + validation + help text.

### DIFF
--- a/src/components/ProfilePage/Bio.jsx
+++ b/src/components/ProfilePage/Bio.jsx
@@ -56,39 +56,12 @@ class Bio extends React.Component {
           editing: (
             <div role="dialog" aria-labelledby={`${formId}-label`}>
               <form onSubmit={this.handleSubmit}>
-                {/*
-                  If we wrote our own form group component, maybe
-                  it would look like this. Encapsulates everything
-                  inside div.form-group
-
-                  <FormGroup
-                    // group info, or shared among child nodes
-                    id={formId}
-                    inline
-                    invalid={error != null}
-
-                    // input info
-                    value={bio}
-                    inputType="textarea"
-                    inputName={formId}
-
-                    // Meta/help/errors
-                    label={intl.formatMessage(messages['profile.bio.about.me'])}
-                    helpText=""
-                    errorMessage={error}
-                    validMessage={successMessage}
-
-                    // event handlers for input
-                    onInputChange={this.handleChange}
-                  />
-                */}
                 <div className="form-group">
-                  <label htmlFor={formId} id={`${formId}-label`}>
+                  <label htmlFor={formId}>
                     {intl.formatMessage(messages['profile.bio.about.me'])}
                   </label>
                   <textarea
                     className={classNames('form-control', { 'is-invalid': Boolean(error) })}
-                    type="textarea"
                     id={formId}
                     name={formId}
                     value={bio}
--- a/src/components/ProfilePage/Bio.jsx
+++ b/src/components/ProfilePage/Bio.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
 
 import messages from './Bio.messages';
 
@@ -86,7 +87,7 @@ class Bio extends React.Component {
                     {intl.formatMessage(messages['profile.bio.about.me'])}
                   </label>
                   <textarea
-                    className={`form-control ${error ? 'is-invalid' : ''}`}
+                    className={classNames('form-control', { 'is-invalid': Boolean(error) })}
                     type="textarea"
                     id={formId}
                     name={formId}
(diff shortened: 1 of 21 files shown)


src/components/ProfilePage/Bio.jsx
```diff
@@ -55,30 +54,56 @@ class Bio extends React.Component {
...
+                  />
+                */}
+                <div className="form-group">
+                  <label htmlFor={formId} id={`${formId}-label`}>
                     {intl.formatMessage(messages['profile.bio.about.me'])}
-                  </Label>
-                  <Input
+                  </label>
+                  <textarea
+                    className={`form-control ${error ? 'is-invalid' : ''}`}
```
davidjoy: Might use `classNames` here?

src/components/ProfilePage/Bio.jsx
```diff
@@ -55,30 +54,56 @@ class Bio extends React.Component {
...
                     name={formId}
-                    value={bio || ''}
-                    invalid={error != null}
+                    value={bio}
                     onChange={this.handleChange}
                     aria-describedby={`${formId}-error-feedback`}
                   />
-                  <FormFeedback id={`${formId}-error-feedback`}>{error}</FormFeedback>
-                </FormGroup>
+                  <p className="invalid-feedback" id={`${formId}-error-feedback`}>{error}</p>
```
davidjoy: One interesting thing - FormFeedback was actually obscuring that this was just a paragraph tag.

src/components/ProfilePage/Name.jsx
```diff
@@ -72,19 +71,19 @@ class Name extends React.Component {
                   such to fully get rid of it.
                   */}
                   <p className="h5">{name}</p>
-                  <FormText id={`${formId}-help-text`}>
+                  <small className="form-text text-muted" id={`${formId}-help-text`}>
```
davidjoy: So this one is interesting, as it's sorta difficult for folks to know we use `small` to represent form helper text.  That said, it's what they do in the bootstrap documentation.  I guess if we had our own FormRow/FormGroup component that'd help with this.  &shrug;

abutterworth: Yea I think a higher level component would manage this. Also, it seems weird to have small here. I thought it was an inline element by default. I would rather moving forward do `<p className="form-text text-muted small">stuff</p>`. Or just update the style for form-text to include the others

src/components/ProfilePage/Bio.jsx
abutterworth: Yea good call – will add

src/components/ProfilePage/ProfileAvatar.jsx
```diff
@@ -64,49 +64,51 @@ class ProfileAvatar extends React.Component {
...
+              description="Upload photo button"
+            />
+          }
+        />
       );
     }
 
     return (
       <Dropdown
-        isOpen={this.state.dropdownOpen}
```
davidjoy: What replaced these two props on the old Dropdown?  `isOpen` and `toggle`?  Can we remove `this.state.dropdownOpen` and `this.toggleDropdown` now?

src/components/ProfilePage/Bio.jsx
```diff
@@ -55,30 +55,56 @@ class Bio extends React.Component {
...
+                */}
+                <div className="form-group">
+                  <label htmlFor={formId} id={`${formId}-label`}>
                     {intl.formatMessage(messages['profile.bio.about.me'])}
-                  </Label>
-                  <Input
+                  </label>
+                  <textarea
+                    className={classNames('form-control', { 'is-invalid': Boolean(error) })}
                     type="textarea"
```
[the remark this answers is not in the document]
abutterworth: As discussed, looks like we're good here: https://reactjs.org/docs/forms.html#the-textarea-tag

src/components/ProfilePage/Bio.jsx
```diff
@@ -55,30 +55,56 @@ class Bio extends React.Component {
...
                     id={formId}
                     name={formId}
-                    value={bio || ''}
-                    invalid={error != null}
+                    value={bio}
```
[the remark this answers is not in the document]
abutterworth: As discussed, it looks like we're good here. This is a react specific thing: https://reactjs.org/docs/forms.html#the-textarea-tag

src/components/ProfilePage/ProfileAvatar.jsx
abutterworth: Removed!

PR #131, opened 2019-04-04
Attempt to log the entire error object on request errors.

douglashall:


src/services/LoggingService.js
```diff
@@ -53,15 +53,14 @@ class LoggingService {
...
       const requestAttributes = {
         errorType: 'api-request-error',
         errorStatus: request.status,
-        errorUrl: requestUrl,
+        errorUrl: request.responseURL,
         errorData: errorMessage,
       };
       updatedCustomAttributes = Object.assign({}, requestAttributes, customAttributes);
-      processedError = new Error(`API request failed: ${request.status} ${requestUrl} ${errorMessage}`);
+      processedError = new Error(`API request failed: ${request.status} ${request.responseURL} ${errorMessage} ${JSON.stringify(error)}`);
```
robrap: @douglashall: You could add it as a custom metric using `fixErrorLength(JSON.stringify(error))` to get as much as possible.

douglashall: Does it matter?

robrap: If it is temporary, it doesn't matter much.  You'll get a few more characters if it is big and you separate it out.

PR #135, opened 2019-04-08
ARCH-517: prepare analytics for move to separate repo

robrap:
- Make analytics and segment files configurable from outside.
- Add checks for previously called identify on page/track.
- Add some additional parameters to tracking events.

ARCH-517

FYI: @nasthagiri

### DIFF
--- a/src/analytics/analytics.js
+++ b/src/analytics/analytics.js
@@ -1,54 +1,133 @@
 import formurlencoded from 'form-urlencoded';
-import LoggingService from '@edx/frontend-logging';
-
-import apiClient from '../config/apiClient';
-import { configuration } from '../config/environment';
 import { snakeCaseObject } from '../services/utils';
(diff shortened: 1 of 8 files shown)


src/analytics/analytics.js
```diff
@@ -1,53 +1,133 @@
...
+  const loggingService = getLoggingService(); // verifies configuration early
+  if (!hasIdentifyBeenCalled) {
+    loggingService.logError('Identify must be called before other tracking events.');
+  }
+}
+
+/**
+ * Logs events to tracking log and downstream.
+ *
+ * Note: these events don't go to Segment unless specifically whitelisted.
```
albemarle: Where does that whitelisting happen?

robrap: I pointed to docs for this instead.

PR #136, opened 2019-04-08
Upgrade and use Paragon

abutterworth:
- Upgrade to Paragon v4
- Change usage of Hyperlink and Button components
- Use ValidationFormGroup for handling validation classes and messages.
- Use StatefulButton instead of local AsyncActionButton
- Use replace TransitionReplace with Paragon's version


src/components/ProfilePage/elements/FormControls.jsx
```diff
@@ -30,20 +29,19 @@ function FormControls({
         />
       </div>
       <div className="form-group">
-        <AsyncActionButton
+        <StatefulButton
```
albemarle: :-)

PR #142, opened 2019-04-10
Reuse userAccount data for profile if possible.

davidjoy:
If we're looking at the profile of the authenticated user, skip calling fetchAccount a second time to load profile data and just re-use the user account data we loaded as part of frontend-auth's work.

It may be useful to review this by commit.  The first commit removes some data massaging we were doing that was ultimately getting in the way.  Before this PR, we were massaging data for the profile to change "levelOfEducation" to "education" as a nicety, but now we want to reuse userAccount data as the profile, that conversion is kind of a pain/was just going to make the process more complicated for little benefit.  So I removed it.  

The second commit relies on a change to frontend-auth to put a `loaded` flag in the userAccount part of the redux store state.  We use this to guarantee that the userAccount is loaded before we decide whether or not we should fetch the profile.  Technically for other users' profiles this makes our requests more serial, but I couldn't find a way to keep them fully parallel without a more significant refactor.


src/components/App.jsx
```diff
@@ -24,41 +25,51 @@ class App extends Component {
     this.props.fetchUserAccount(userAccountApiService, username);
   }
 
+  renderContent() {
+    if (!this.props.loaded) {
```
davidjoy: Pulling this out into a separate function and conditionally returning on `loaded` allows us to defer calling `fetchProfile` in ConnectedProfilePage until we've gotten the userAccount data.  This is the serialization I was talking about in the PR description.  We don't even want to show the ConnectedProfilePage if the user isn't authenticated, so I didn't want to bother making requests for the profile data here in componentDidMount() before we know if we even need to.

src/sagas/RootSaga.js
```diff
@@ -38,29 +38,39 @@ import { keepKeys } from '../services/utils';
 
 export function* handleFetchProfile(action) {
   const { username } = action.payload;
-  const currentUsername = yield select(state => state.authentication.username); // eslint-disable-line
+  const { isCurrentUserProfile, userAccount } = yield select(handleFetchProfileSelector);
+
+  // Default our data.
+  let preferences = {};
+  let account = userAccount;
```
davidjoy: Assume we'll use the current user account data for the page - this is overridden below if we're not looking at the current user's profile.

src/selectors/ProfilePageSelector.js
```diff
@@ -164,12 +165,21 @@ export const handleSaveProfileSelector = createSelector(
...
+  }),
+);
+
 // Reformats the social links in a platform-keyed hash.
 const socialLinksByPlatformSelector = createSelector(
   profileAccountSelector,
   (account) => {
     const linksByPlatform = {};
-    if (account.socialLinks !== undefined) {
+    if (account.socialLinks != null) {
```
davidjoy: The `userAccount` reducer has different defaults for socialLinks, so this needed to be made a bit less stringent, along with the one below.  In our profile reducer, we default it to an empty array, whereas in userAccount it's just null.

src/sagas/RootSaga.js
```diff
@@ -38,29 +38,39 @@ import { keepKeys } from '../services/utils';
 
 export function* handleFetchProfile(action) {
   const { username } = action.payload;
-  const currentUsername = yield select(state => state.authentication.username); // eslint-disable-line
+  const { isCurrentUserProfile, userAccount } = yield select(handleFetchProfileSelector);
```
douglashall: Nit: `isAuthenticatedUserProfile`

src/sagas/RootSaga.js
```diff
@@ -38,29 +38,39 @@ import { keepKeys } from '../services/utils';
...
+
+  // Default our data.
```
albemarle: I don't think this comment is necessary.

src/sagas/RootSaga.js
```diff
@@ -38,29 +38,39 @@ import { keepKeys } from '../services/utils';
...
+    // Depending on which profile we're loading, we need to make different calls.
     const calls = [
-      call(ProfileApiService.getAccount, username),
+      // We'll always make a call for certificates.
       call(ProfileApiService.getCourseCertificates, username),
     ];
 
-    if (username === currentUsername) {
+    if (isCurrentUserProfile) {
+      // If the profile is for the current user, get their preferences.
```
albemarle: Same here.  Ideally what's happening should be clear from the code (which it is here); comments should be for when you have to add context to explain _why_ you're doing something.

src/selectors/ProfilePageSelector.js
albemarle: Should it become a function?  `hasSocialLinks` or similar?

src/sagas/RootSaga.js
davidjoy: I add comments when I feel like the decisions in a piece of code are opaque to someone without any context.  In this function's case, there are specific calls we need to make in different situations, which interacts with how we're setting up the data.  I'm okay with some hand holding if it helps anyone wondering what's going on, personally.  That said, I'm updating the comments to be more specific to the decision process.

src/sagas/RootSaga.js
davidjoy: Doing this. 👍

src/selectors/ProfilePageSelector.js
davidjoy: Eh... could be, but I don't think it'll see much reuse or anything.

davidjoy: Actually, I'm going to change it to use `Array.isArray()` - that'll be more specific to what this is trying to check for.

albemarle: I thought you had to change it in two places in this PR-- or is line 195 a different object?

PR #143, opened 2019-04-10
ARCH-517: use new frontend-analytics library

robrap:
Moves analytics code to new frontend-analytics library.

ARCH-517

Before Merge:
- [x] Update package and package-lock to point to new npm package.


package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "@cospired/i18n-iso-languages": "^2.0.2",
     "@edx/edx-bootstrap": "git://github.com/edx/edx-bootstrap.git#update-with-documentation-site",
+    "@edx/frontend-analytics": "file:../frontend-analytics",
```
robrap: Reviewers: This will be updated before merge.

davidjoy: Related to this, maybe, are those 15k lines added to package-lock.json... something seems wrong there. Looking at the definition of `@edx/frontend-auth` in that file, it's only like a dozen lines long, whereas the definition for frontend-analytics is... 15k lines. :)

robrap: Yes.  Should be all set when I can switch to the real package.

robrap: FYI: this is all set.

PR #148, opened 2019-04-10
feat: lookup language preference in cookie

abutterworth:
Some notes:

- Adds a dependency `universal-cookie`.
- the cookie value is a two-letter language code
- Should not be merged until stage and prod configuration is updated

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -144,6 +144,7 @@ module.exports = Merge.smart(commonConfig, {
       ACCESS_TOKEN_COOKIE_NAME: null,
       USER_INFO_COOKIE_NAME: null,
       CSRF_COOKIE_NAME: 'csrftoken',
+      LANGUAGE_PREFERENCE_COOKIE_NAME: null,
       NEW_RELIC_APP_ID: null,
       NEW_RELIC_LICENSE_KEY: null,
       SITE_NAME: null,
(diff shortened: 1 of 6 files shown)


config/webpack.prod.config.js
```diff
@@ -144,6 +144,7 @@ module.exports = Merge.smart(commonConfig, {
       ACCESS_TOKEN_COOKIE_NAME: null,
       USER_INFO_COOKIE_NAME: null,
       CSRF_COOKIE_NAME: 'csrftoken',
+      LANGUAGE_PREFERENCE_COOKIE_NAME: null,
```
albemarle: Really?

src/i18n/i18n-loader.js
```diff
@@ -46,9 +47,20 @@ LANGUAGES.registerLocale(require('@cospired/i18n-iso-languages/langs/en.json'));
...
+const cookies = new Cookies();
+
+// Get the locale by setting priority
+const getLocale = (localeStr) => {
+  // 1. Explicit application request
+  if (localeStr) return localeStr;
+
+  // 2. User setting in cookie
+  const cookieLanguagePreference = cookies.get(process.env.LANGUAGE_PREFERENCE_COOKIE_NAME);
+  if (cookieLanguagePreference) return cookieLanguagePreference;
```
albemarle: We should probably check if we actually offer that language... maybe even for #1 above.

config/webpack.prod.config.js
abutterworth: They get overridden by edx-internal variables (not sure exactly what this process is)

robrap: I imagine it will need to get added to the prod/stage files here:
https://github.com/edx/edx-internal/blob/29a43bc351b18621ff8dbe3d54dec0103d0619a5/frontends/common/prod_config.yml

abutterworth: We're all set there

PR #151, opened 2019-04-10
fix: get two letter language code from cookie

abutterworth:
Can now handle values like `es-415` in the cookie


src/i18n/i18n-loader.js
```diff
@@ -65,7 +65,7 @@ const getLocale = (localeStr) => {
   // 2. User setting in cookie
   const cookieLanguagePreference = cookies.get(process.env.LANGUAGE_PREFERENCE_COOKIE_NAME);
   if (cookieLanguagePreference && messages[cookieLanguagePreference] !== undefined) {
-    return cookieLanguagePreference;
+    return cookieLanguagePreference.substr(0, 2);
```
douglashall: You could break this out into a `parseLanguageCode` function now that you are doing it twice. Or not.

PR #153, opened 2019-04-11
Organize application into modules.

davidjoy:
Related article: https://jaysoo.ca/2016/02/28/organizing-redux-application/

The principles:

1. Code is grouped into modules.  Modules can be nested as necessary, though that wasn't really needed here.
2. Code in a module does not import from its parent module, as if it were an npm dependency of the parent.  
3. A module should conceptually be able to be extracted into its own npm package with minimal effort.
4. Likewise, a more general module shouldn't be importing from a more specific one (i.e., the "common" module shouldn't be importing from the "profile" module, but the reverse is okay.  If two modules need a shared resource, it needs to be bumped up a level in generality.
5. Modules are configured from above.  For react components, this is giving them props, for services, this is giving it apiClient/configuration, etc.
6. A module has a public interface, as specified in its top level index.js file.
7. Code that imports a module should only use things exposed in the public interface/shouldn't import from deeper down in the module.
8. Inside a module directory, there should be one file for each type of code that it uses: reducers.js, sagas.js, actions.js, services.js, etc.  Components go in a components directory, and if more than one file of the other types are necessary, they can be put into - for example - a "reducers" directory and then named semantically according to their role.  reducers/FormReducers.js, reducers/OtherReducers.js, etc.  
9. If a clear sub-module emerges, it can be put into a sub-directory with its own set of files for each type of code.  

Another structure I haven't tried: https://hackernoon.com/fractal-a-react-app-structure-for-infinite-scale-4dab943092af

And if we do decide we like this layout, I'd like to ultimately write it up as an... OEP, I think.  Either that or some sort of blog post, because I feel like there's a lot of good things that happened as a result of it, and there are certain rules to observe.

# Overview

There are effectively three loosely coupled "modules" here:

1. The micro-frontend boilerplate.
2. The profile reusable organism. 
3. A module of "common" code.  This is stuff that's not specific to the MFE or to the profile, but that might be used by both or otherwise belong outside this repo/want to be reused.  

Doing this rearrangement causes the directories to be semantically meaningful and the files to be named fairly generically.  Each module (common and profile) has a small set of exports, and is otherwise a black box.  In theory, the profile/common modules could be ripped out and put into their own repositories fairly easily.  This means that the MFE boilerplate code should be nearly identical for all MFEs now.

In general, I observed that both relative paths and filenames got a lot shorter in general.  Code that works together is also, in general, much closer together.  Most things are merely a `"./"`away.

In addition to moving files around, there were a few things I tried to do:

1. Configure the profile organism instead of having it import configuration it needed.  This is related to `apiClient` and `configuration` primarily.  The profile organism uses some things from common, but it should not reach outside its own directory, otherwise.
2. Simplify the configuration of SiteHeader, make it more like SiteFooter.
3. The MFE boilerplate and profile module respect the boundaries of modules and only import from that module's top level index.js file (it's public interface).
4. Gather all the app configuration at the top, in src/index.jsx.  This isn't the prettiest, and some of it might be better shuffled off into a separate file, but the exercise made it really clear what depended on what.


src/common/actions.js
```diff
@@ -0,0 +1,11 @@
+import { fetchUserAccount as _fetchUserAccount, UserAccountApiService } from '@edx/frontend-auth';
```
davidjoy: So this file is of interest.  These are some acrobatics to hide the existence of UserAccountApiService and apiClient from the profile module.  As you can see here, the default implementation of `fetchUserAccount` needs an instance of `UserAccountApiService` to be passed to it, which then in turn needs the configured `apiClient`.  By doing this and exposing a "wrapped" `fetchUserAccount`, we can have the high-level MFE boilerplate configure the service and the organisms in it are none-the-wiser about what it took to set it up.  This helps support the modularity of the profile.

src/common/components/PageLoading.jsx
```diff
@@ -1,11 +1,8 @@
 import React from 'react';
 
-import Banner from './Banner';
```
davidjoy: Banner is profile-specific, but PageLoading isn't.  I wanted to use PageLoading in the MFE boilerplate, so I split out Banner and put it in place in the profile while it's loading.  

This isn't perfect, as we probably want the banner showing during _all_ preloaders, but that'd be a deeper refactor of our data loading in this app.

src/common/index.js
```diff
@@ -0,0 +1,10 @@
+import * as utils from './utils';
```
davidjoy: So this "common" module can nearly be subsumed by other, existing modules:

1. PageLoading, if reusable, probably belongs in Paragon.
2. utils are useful for APIs, and probably belong in either frontend-auth (as the place we setup our API clients) or some sort of frontend-api library.
3. As mentioned above, configureUserAccountApiService and fetchUserAccount probably just belong in frontend-auth ultimately.

That means this module doesn't even need to exist!  If this were a more complicated MFE with multiple pages (read: modules), then we might bring this back for code that is shared across them but is otherwise specific to the MFE.

src/common/utils.js
```diff
@@ -46,3 +46,36 @@ export function keepKeys(data, whitelist) {
...
   return result;
 }
+
+/**
+ * Helper class to save time when writing out action types for asynchronous methods.  Also helps
+ * ensure that actions are namespaced.
+ *
+ * TODO: Put somewhere common to it can be used by other MFEs.
+ */
+export class AsyncActionType {
```
davidjoy: I had this class in a separate file - I just moved it in here to simplify.

src/components/App.jsx
```diff
@@ -1,75 +1,180 @@
...
 
-class App extends Component {
-  componentDidMount() {
-    const { username } = this.props;
-    const userAccountApiService = new UserAccountApiService(apiClient, process.env.LMS_BASE_URL);
-    this.props.fetchUserAccount(userAccountApiService, username);
+import messages from './App.messages';
+
+
+function PageContent({
```
davidjoy: So this PageContent component is a bit of a compromise.  I wanted to simplify the SiteHeader configuration, but it _does_ need a reference to `intl`.  Since, originally, the SiteHeader was rendered in our top-level component where we were also setting up `IntlProvider`, the `intl` variable didn't exist yet!  But by breaking this part out into its own component, the top-level App component will setup its providers prior to PageContent being rendered.  Thus, `intl` will be available.

Part of what I wanted to do here was, again, get the SiteHeader and SiteFooter configuration all in one place and separated from redux so we can visualize what it might look like if it were passed in from some outside source (netlify, a .json file, etc.)

src/components/App.jsx
```diff
@@ -1,75 +1,180 @@
...
+        avatar={avatar}
+        logoAltText={configuration.SITE_NAME}
+        logoDestination={configuration.MARKETING_SITE_BASE_URL}
+        mainMenu={mainMenu}
+        userMenu={userMenu}
+        loggedOutItems={loggedOutItems}
+      />
+      <main>
+        <Switch>
+          <Route path="/u/:username" component={ConnectedProfilePage} />
```
davidjoy: In a larger MFE, and if the reusable organisms were actually in their own repos/own npm packages, you could imagine replacing this statically-defined `<Switch>` with something that discovers the pages it wants to include.  Plugin-architecture-style.  Doing that would help make App.jsx even more de-coupled from the actual content of a particular MFE, and thus more reusable.

src/components/App.jsx
```diff
@@ -1,75 +1,180 @@
...
+  }).isRequired,
+  intl: intlShape.isRequired,
+};
+
+PageContent.defaultProps = {
+  ready: false,
+  avatar: null,
+};
+
+const IntlPageContent = injectIntl(PageContent);
```
davidjoy: The magic line that required the refactoring mentioned above - separating out PageContent lets us use `injectIntl` here.

src/components/App.jsx
```diff
@@ -80,20 +185,43 @@ class App extends Component {
...
   ready: false,
+  avatar: null,
 };
 
 const mapStateToProps = state => ({
   username: state.authentication.username,
   // An error means that we tried to load the user account and failed,
   // which also means we're ready to display something.
   ready: state.userAccount.loaded || state.userAccount.error != null,
+  configuration: state.configuration,
```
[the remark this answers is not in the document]
davidjoy: Also, I _think_ it's a good idea not to use `process.env.*` in the app and to abstract to this "configuration" object.  It lets us be agnostic to where the configuration is actually coming from.  And in truth, a bunch of our config vars are derived from stuff in process.env anyway/need a home.

src/components/ErrorPage.jsx
```diff
@@ -1,46 +1,51 @@
...
 import { FormattedMessage } from 'react-intl';
 import { Link } from 'react-router-dom';
 
-import apiClient from '../config/apiClient';
-
-export default class ErrorPage extends Component {
-  componentDidMount() {}
-
-  render() {
-    const { username } = apiClient.getAuthenticationState().authentication;
```
davidjoy: Since this authentication state is in redux, I removed this dependency on apiClient and just got it from there instead (by connecting this component).

src/environment.js
```diff
@@ -1,18 +1,34 @@
...
   CSRF_TOKEN_API_PATH: process.env.CSRF_TOKEN_API_PATH,
   REFRESH_ACCESS_TOKEN_ENDPOINT: process.env.REFRESH_ACCESS_TOKEN_ENDPOINT,
-  DATA_API_BASE_URL: process.env.DATA_API_BASE_URL,
-  SECURE_COOKIES: process.env.NODE_ENV !== 'development',
   SEGMENT_KEY: process.env.SEGMENT_KEY,
   ACCESS_TOKEN_COOKIE_NAME: process.env.ACCESS_TOKEN_COOKIE_NAME,
   USER_INFO_COOKIE_NAME: process.env.USER_INFO_COOKIE_NAME,
   CSRF_COOKIE_NAME: process.env.CSRF_COOKIE_NAME,
   LANGUAGE_PREFERENCE_COOKIE_NAME: process.env.LANGUAGE_PREFERENCE_COOKIE_NAME,
+  SITE_NAME: process.env.SITE_NAME,
```
davidjoy: So all these were missing.

src/index.jsx
```diff
@@ -1,30 +1,70 @@
...
-import apiClient from './config/apiClient';
+import { configuration } from './environment';
 import { handleRtl } from './i18n/i18n-loader';
+import configureStore from './store';
+import { configureProfileApiService } from './profile';
+import { configureUserAccountApiService } from './common';
 
 import './index.scss';
-
 import App from './components/App';
```
davidjoy: Okay!  So I added stuff here.  This file is ugly and could be factored better, but that fact aside, this is where the magic happens now.  

1. apiClient is now created here, first thing, because nearly everything depends on it.  Rather than it being imported here and there and everywhere, it's passed into code that depends on it via configuration.  
2. After apiClient has determined we're allowed to be here (via `ensurePublicOrAuthenticationAndCookies`), we then configure everything else we need.  The redux store, services, segment, analytics, and RTL.  They all depend on some combination of the apiClient and our configuration, so we hand those two to them here.  Prior to this, they all just imported one or both, which created a hard dependency on those files knowing that there was a file called "apiClient.js" or "environment.js".  

Again, this could be cleaned up and improved, but gathering it all together here means each of these things is no longer reaching out to find the config/apiClient themselves.

src/index.jsx
```diff
@@ -1,30 +1,70 @@
...
+
 apiClient.ensurePublicOrAuthenticationAndCookies(
   window.location.pathname,
   () => {
-    const { store, history } = configureStore();
-
-    if (process.env.NODE_ENV === 'production') {
-      handleRtl();
-    }
+    const { store, history } = configure();
```
davidjoy: I put everything into `configure` as a token effort at organization. ;)

src/profile/components/AgeMessage.jsx
```diff
@@ -1,12 +1,9 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { StatusAlert } from '@edx/paragon';
 import { FormattedMessage } from 'react-intl';
 
-import { configuration } from '../../config/environment';
-
-const { ACCOUNT_SETTINGS_URL } = configuration;
```
davidjoy: Example of the tight coupling we had before.  AgeMessage should definitely not know there's some random file outside its module called environment.js where it can get a URL it needs.

src/profile/components/ProfilePage.jsx
```diff
@@ -16,25 +14,22 @@ import {
...
-import Bio from './ProfilePage/Bio';
-import Certificates from './ProfilePage/Certificates';
-import AgeMessage from './ProfilePage/AgeMessage';
-import DateJoined from './ProfilePage/DateJoined';
-import PageLoading from './common/PageLoading';
-import Banner from './common/Banner';
-import { profilePageSelector } from '../selectors/ProfilePageSelector';
-
-// Configuration
-import { configuration } from '../config/environment';
```
davidjoy: Again - goodbye import!

src/profile/components/ProfilePage.jsx
```diff
@@ -155,98 +161,108 @@ export class ProfilePage extends React.Component {
...
+        <Banner />
+        {this.renderContent()}
+      </div>
+    );
+  }
 }
 
 ProfilePage.propTypes = {
+  // Config
+  configuration: PropTypes.shape({
```
davidjoy: The profile page, without knowing where the config came from, can identify some URLs that it knows it needs.  This will yell and scream in development if those haven't been properly configured in its parent app, but it doesn't get to know anything about where they actually came from.

src/profile/components/__mocks__/loadingApp.mockStore.js
```diff
@@ -3,6 +3,10 @@ module.exports = {
     userId: 9,
     username: 'staff'
   },
+  configuration: {
```
davidjoy: Ideally these "mockStore" files should probably be replaced with factories... this is pretty un-DRY.

src/profile/constants.js
```diff
@@ -0,0 +1,28 @@
+const EDUCATION_LEVELS = [
```
davidjoy: Consolidated all these constants into one file for simplicity.

src/profile/index.js
```diff
@@ -0,0 +1,11 @@
+import profileReducer from './reducers';
+import profileSaga from './sagas';
+import ConnectedProfilePage from './components/ProfilePage';
+import configureProfileApiService from './services';
+
+export {
```
davidjoy: So this is cool!  The whole profile module has four things it exports.  That's its entire interface with the outside world.

davidjoy: This means that to use the profile module, you just import from this file, rather than importing from `./profile/<anything else>`.  Technically there isn't a good way to enforce that, that I'm aware of, short of moving the profile organism into its own repo, though.

davidjoy: Also, the profile module should not be reaching outside of the `profile` directory... ever!  The one exception right now is that it uses things in the 'common' module, but as mentioned above, all of those should be shuffled off into other repos/npm modules, and/or it represents common code that multiple modules may need.  Otherwise, `profile` doesn't reach up into the MFE boilerplate at all now.

src/profile/services.js
```diff
@@ -1,12 +1,32 @@
...
+
+function validateConfiguration(newConfig) {
+  Object.keys(config).forEach((key) => {
+    if (newConfig[key] === undefined) {
+      throw new Error(`Service configuration error: ${key} is required.`);
+    }
+  });
+}
+
+export default function configure(newConfig, newApiClient) {
```
davidjoy: So this is new - it's where you give the profile API service the data it needs.  It then validates that you've configured it properly.

src/reducers.js
```diff
@@ -14,8 +14,9 @@ const createRootReducer = history =>
     // The authentication state is added as initialState when
     // creating the store in data/store.js.
     authentication: identityReducer,
+    configuration: identityReducer,
```
davidjoy: The configuration data was added to redux so connected components have access to it without knowing how it got there.

src/reducers.js
```diff
@@ -14,8 +14,9 @@ const createRootReducer = history =>
...
     userAccount,
-    profilePage,
+    profilePage: profileReducer,
```
davidjoy: In a world where profilePage was a plugin and/or we wanted to have a common MFE boilerplate module that was shared across MFEs, you could imagine we do something fancy to remove this direct dependency on the profile reducer.  I'm not sure what that looks like, but it's basically part of a plugin architecture for MFEs.

src/sagas.js
```diff
@@ -0,0 +1,9 @@
+import { all } from 'redux-saga/effects';
+
+import { profileSaga } from './profile';
+
+export default function* rootSaga() {
+  yield all([
+    profileSaga(),
```
davidjoy: Another touch point between the MFE boilerplate and profile module - like reducers above.

src/store/configureStore.dev.js
```diff
@@ -6,18 +6,16 @@ import { routerMiddleware } from 'connected-react-router';
...
 import { createLogger } from 'redux-logger';
 
-import apiClient from './apiClient';
-import createRootReducer from '../reducers/RootReducer';
-import rootSaga from '../sagas/RootSaga';
+import createRootReducer from '../reducers';
+import rootSaga from '../sagas';
 
-export default function configureStore() {
+export default function configureStore(initialState = {}) {
```
davidjoy: This is a more canonical way of getting initial state into the store.  That way the store configuration doesn't know anything about apiClient/where data came from.

src/store/index.js
```diff
@@ -0,0 +1,9 @@
+import configureStoreProd from './configureStore.prod';
+import configureStoreDev from './configureStore.dev';
+
+export default function configureStore(state, env) {
```
davidjoy: Hm, this file was a rename from configureStore.js; apparently it lost that.  (There wasn't much history in there).  Anyway, we're passing in the `env` here so that we don't have to use `process.env` or `configuration` directly.

webpack/webpack.prod.config.js
```diff
@@ -162,6 +160,8 @@ module.exports = Merge.smart(commonConfig, {
       REDDIT_URL: null,
       APPLE_APP_STORE_URL: null,
       GOOGLE_PLAY_URL: null,
+      NEW_RELIC_APP_ID: null,
```
davidjoy: Put these at the bottom as they're not in the dev version/I wasn't sure if they should be.

albemarle:
image

src/components/App.jsx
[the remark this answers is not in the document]
robrap: What about React Context?  I'm not saying this PR needs to change, just food for thought.

src/index.jsx
robrap: Not looking at code, but just your comments.  We just need to ensure we don't lose the identify/page event timing that comes after getting user info.

src/components/App.jsx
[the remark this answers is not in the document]
davidjoy: Context is kind of hard to use, from what I've seen; I initially tried that.  `injectIntl` basically makes it easier by dumping the intl context into a component's props.

davidjoy:
Note: I think we'll want to do a pretty hefty regression test on this branch prior to deploying to prod.  Obviously, it touched a lot of code and it's entirely possible there's a little typo or something somewhere!

src/profile/index.js
```diff
@@ -0,0 +1,11 @@
+import profileReducer from './reducers';
+import profileSaga from './sagas';
+import ConnectedProfilePage from './components/ProfilePage';
+import configureProfileApiService from './services';
```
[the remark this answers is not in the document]
abutterworth: And maybe even more generically in ./src/index.jsx:
```
import { configure as configureProfile } from './profile';

configureProfile({
  apiService: { configuration, apiClient },
  // ... other stuff for in the future or different kinds of modules
});

src/common/actions.js
davidjoy: It'd be nice if this code didn't have to exist, though.

src/components/App.jsx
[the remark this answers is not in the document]
davidjoy: Agreed - it seemed kinda onerous, so I punted on figuring it out.

src/profile/components/__mocks__/loadingApp.mockStore.js
davidjoy: It'd help make sure they don't get out of date by having something like:

factory.create('configuration'), where 'configuration' was a kind of known object with a definition in a factory file somewhere.  All tests that need a configuration object would get it from the same factory, so they'd all get updated if it changes, basically.  Same would go for any other object in there, like userAccount.

src/profile/index.js
davidjoy: Mostly?  I'd say if you find yourself looking outside a module for something, think twice.  I kinda fleshed this out in those "principles" in the PR description _after_ I wrote these comments...  Within a module, any time you have some code reuse, that code's likely to be in a "common" sibling directory to the places using it, I'll bet.

src/profile/index.js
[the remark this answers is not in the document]
davidjoy: I think you're right - I had similar thoughts but was trying not to dive TOO deep on getting everything perfect. :)  But yes!

davidjoy:
@abutterworth Agreed on both your points 1 and 2!

PR #155, opened 2019-04-12
fix: make intl.formatMessage more resilient

abutterworth:
- Adds a shim replacement for injectIntl to override intl.formatMessage so it won't blow up when supplied an undefined message definition.
- Adds an index.js file to the i18n folder to start prepping it to be pulled out into a package
- Added an alias for @edx/frontend-i18n in the webpack config and jest config (package.json) to allow the rest of the code be written as if this package already existed.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -9,7 +9,7 @@ import SiteFooter from '@edx/frontend-component-footer';
 import { fetchUserAccount, UserAccountApiService } from '@edx/frontend-auth';
 
 import apiClient from '../config/apiClient';
-import { getLocale, getMessages } from '../i18n/i18n-loader';
+import { getLocale, getMessages } from '@edx/frontend-i18n'; // eslint-disable-line
 import SiteHeader from './common/SiteHeader';
 import ConnectedProfilePage from './ProfilePage';
 
--- a/src/i18n/injectIntlWithShim.jsx
+++ b/src/i18n/injectIntlWithShim.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import { injectIntl, intlShape } from 'react-intl';
+
+
+const injectIntlWithShim = (WrappedComponent) => {
+  class ShimmedIntlComponent extends React.Component {
+    static propTypes = {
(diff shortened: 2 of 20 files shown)


src/components/App.jsx
```diff
@@ -9,7 +9,7 @@ import SiteFooter from '@edx/frontend-component-footer';
 import { fetchUserAccount, UserAccountApiService } from '@edx/frontend-auth';
 
 import apiClient from '../config/apiClient';
-import { getLocale, getMessages } from '../i18n/i18n-loader';
+import { getLocale, getMessages } from '@edx/frontend-i18n'; // eslint-disable-line
```
albemarle: What does it complain about?

src/i18n/injectIntlWithShim.jsx
```diff
@@ -0,0 +1,39 @@
...
+    static propTypes = {
+      intl: intlShape.isRequired,
+    };
+
+    constructor(props) {
+      super(props);
+      this.shimmedIntl = Object.create(this.props.intl, {
+        formatMessage: {
+          value: (definition) => {
+            if (definition === undefined) {
```
albemarle: Can we add a check here to also make sure that it's an object with an id property?

src/i18n/injectIntlWithShim.jsx
```diff
@@ -0,0 +1,39 @@
...
+              if (process.env.NODE_ENV !== 'production') {
+                // eslint-disable-next-line no-console
+                console.error('i18n error: An undefined message was supplied to intl.formatMessage.');
+                return '!!! Missing message supplied to intl.formatMessage !!!';
+              }
+              return ''; // Fail silent in production
```
albemarle: Can we use the error logging code instead?

src/components/App.jsx
abutterworth: eslint can't resolve the import properly. To get it to work you need to add some packages like this one: https://www.npmjs.com/package/eslint-import-resolver-alias

I figured it wasn't worth the effort.

PR #158, opened 2019-04-16
fix: remove use of URL which is unsupported in IE 11

abutterworth:
Our usage of `new URL()` to parse certificate download URLs causes errors in IE 11 where it is unsupported. This refactors the construction of the download URL, which is supplied either as a path or a full URL.

### DIFF
--- a/src/profile/services.js
+++ b/src/profile/services.js
@@ -142,10 +142,19 @@ export async function patchPreferences(username, params) {
 function transformCertificateData(data) {
   const transformedData = [];
   data.forEach((cert) => {
+    // download_url may be full url or absolute path.
+    // note: using the URL() api breaks in ie 11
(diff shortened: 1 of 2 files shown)


src/profile/services.js
```diff
@@ -142,10 +142,19 @@ export async function patchPreferences(username, params) {
 function transformCertificateData(data) {
   const transformedData = [];
   data.forEach((cert) => {
+    // download_url may be full url or absolute path.
+    // note: using the URL() api breaks in ie 11
+    const urlIsPath = typeof cert.download_url === 'string' &&
+      cert.download_url.search(/http[s]?:\/\//) !== 0;
```
albemarle: Does this download link appear only for your own certificates?

PR #159, opened 2019-04-22
fix: loading of certificate images

abutterworth:
- Remove extra loader from webpack for svgs. Was svgr-loader, url-loader, and file-loader. Url-loader and file-loader together were creating problems.
- Add a better key for iterating over certificates. It was downloadURL and now it's the courseId + modifiedDate.
- Removed program certificates from the frontend code for now.

### DIFF
--- a/src/profile/components/__snapshots__/ProfilePage.test.jsx.snap
+++ b/src/profile/components/__snapshots__/ProfilePage.test.jsx.snap
@@ -1135,7 +1135,7 @@ exports[`<ProfilePage /> Renders correctly in various states viewing own profile
                     className="certificate-type-illustration"
                     style={
                       Object {
-                        "backgroundImage": "url(null)",
+                        "backgroundImage": "url([object Object])",
                       }
                     }
                   />
@@ -2201,7 +2201,7 @@ exports[`<ProfilePage /> Renders correctly in various states while saving an edi
                     className="certificate-type-illustration"
                     style={
                       Object {
-                        "backgroundImage": "url(null)",
+                        "backgroundImage": "url([object Object])",
                       }
                     }
                   />
--- a/src/profile/components/forms/Certificates.jsx
+++ b/src/profile/components/forms/Certificates.jsx
@@ -14,7 +14,6 @@ import EditableItemHeader from './elements/EditableItemHeader';
 import SwitchContent from './elements/SwitchContent';
 
 // Assets
-import microMastersSVG from '../../assets/micro-masters.svg';
 import professionalCertificateSVG from '../../assets/professional-certificate.svg';
 import verifiedCertificateSVG from '../../assets/verified-certificate.svg';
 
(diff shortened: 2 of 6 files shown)


webpack/webpack.dev.config.js
```diff
@@ -63,10 +63,7 @@ module.exports = Merge.smart(commonConfig, {
         issuer: {
           test: /\.jsx?$/,
         },
-        exclude: [
```
davidjoy: This exclusion wasn't necessary?

src/profile/components/forms/Certificates.jsx
```diff
@@ -50,24 +49,25 @@ class Certificates extends React.Component {
...
-    const certificateIllustration = ((type) => {
-      switch (type) {
-        case 'Professional Certificate':
+    const certificateIllustration = (() => {
+      switch (certificateType) {
+        case 'professional':
+        case 'no-id-professional':
           return professionalCertificateSVG;
-        case 'MicroMasters Certificate':
-          return microMastersSVG;
```
davidjoy: No more MM SVG?

src/profile/components/__snapshots__/ProfilePage.test.jsx.snap
```diff
@@ -1135,7 +1135,7 @@ exports[`<ProfilePage /> Renders correctly in various states viewing own profile
                     className="certificate-type-illustration"
                     style={
                       Object {
-                        "backgroundImage": "url(null)",
+                        "backgroundImage": "url([object Object])",
```
davidjoy: Is it weird that it's dumping an object into the URL for the snapshot?

webpack/webpack.dev.config.js
abutterworth: It didn't seem so. I figured I should remove it if it wasn't having an impact anymore. It could be that this was added to fix the extra loader problem that the certificate images were also having.

src/profile/components/forms/Certificates.jsx
abutterworth: We don't currently display any program certificates. At the moment this case would never be used.

src/profile/components/__snapshots__/ProfilePage.test.jsx.snap
abutterworth: I think this is the normal behavior for the result of a loaded file? I'd bet this is jest's way of handling unknown webpack loaders.

albemarle:
I couldn't see in this PR, do we order certificates, or just display them as we get them from the API?

abutterworth:
I don't think we do any client-side sorting, so whatever order the API
returns them in is what we display right now.

src/profile/components/__snapshots__/ProfilePage.test.jsx.snap
davidjoy: That might make sense - k.

webpack/webpack.dev.config.js
davidjoy: Guess we'll see on staging. :)

PR #161, opened 2019-04-23
Miscellaneous updates to the boilerplate code.

davidjoy:
Based on changes/things I noticed when putting together the frontend-app-account repository.

- Adding ADR stub
- Bumping a few package.json versions.
- Adding a screen reader message to PageLoading component.
- Using LMS_BASE_URL to get the profile link (so it goes through the redirect)
- Adding ECOMMERCE_BASE_URL for consistency with frontend-app-ecommerce config.


webpack/webpack.dev.config.js
```diff
@@ -109,6 +109,7 @@ module.exports = Merge.smart(commonConfig, {
       BASE_URL: 'localhost:1995',
       LMS_BASE_URL: 'http://localhost:18000',
       CREDENTIALS_BASE_URL: 'http://localhost:18150',
+      ECOMMERCE_BASE_URL: 'http://localhost:18130',
```
douglashall: Why add this if we aren't going to use it here?

davidjoy: Considered that - I added it under the assumption that eventually this config setup will be coming from a common place and that it'd be convenient if they were all the same when that DRYing happened.

PR #162, opened 2019-04-24
ARCH-587: upgrade newrelic sourcemap dependency

robrap:
temporarily using a fork to determine how best to proceed.

ARCH-587

Although the Source Maps are loading into NewRelic, they aren't working correctly yet.  Since they aren't working properly, I want to discuss with NewRelic Support, but I imagine the first thing they will want to know is if we are using their latest library.  So, this makes use of their latest.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11380,8 +11380,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -11402,14 +11401,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -15834,12 +15815,11 @@
...
     "new-relic-source-map-webpack-plugin": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/new-relic-source-map-webpack-plugin/-/new-relic-source-map-webpack-plugin-1.1.0.tgz",
-      "integrity": "sha512-evLeCEU6T2ir5ZWN79q2n9Rl94jv62O6XQvhl2vilsFDrUgDu5xzamwXQQtcvySE3gKcO34oozW1xFDp89IaMQ==",
+      "version": "git://github.com/robrap/new-relic-source-map-webpack-plugin.git#c254dfa48de567e43bbeb4389e623081e0674560",
+      "from": "git://github.com/robrap/new-relic-source-map-webpack-plugin.git#update-newrelic-plugin",
       "dev": true,
       "requires": {
-        "@newrelic/publish-sourcemap": "^4.1.2"
+        "@newrelic/publish-sourcemap": "^4.4.1"
```
robrap: FYI: This is the upgrade that I am looking for.

robrap:
Thanks @albemarle.  I think the issue was for sourcemaps of dependencies, so I am going to close this in favor of this other PR: https://github.com/edx/frontend-app-profile/pull/164

PR #165, opened 2019-04-26
ARCH-687: update frontend-logging and frontend-auth

robrap:
- The logging service is now configured and used via an interface,
rather then from the concrete implementation.
- The logging service is passed to frontend-auth.

ARCH-687

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@edx/frontend-auth": "^4.0.0",
     "@edx/frontend-component-footer": "^2.0.3",
     "@edx/frontend-component-site-header": "^2.1.4",
-    "@edx/frontend-logging": "^1.0.2",
+    "@edx/frontend-logging": "^2.0.0",
     "@edx/paragon": "^4.1.3",
     "@fortawesome/fontawesome-svg-core": "^1.2.14",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -27,10 +27,10 @@
     "@cospired/i18n-iso-languages": "^2.0.2",
     "@edx/edx-bootstrap": "^2.0.1",
     "@edx/frontend-analytics": "^1.0.0",
-    "@edx/frontend-auth": "^4.0.0",
+    "@edx/frontend-auth": "file:../frontend-auth",
```
robrap: Reviewers: I will update these once they have landed.

src/i18n/injectIntlWithShim.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { injectIntl, intlShape } from 'react-intl';
-import LoggingService from '@edx/frontend-logging';
+import { logError } from '@edx/frontend-logging';
```
davidjoy: Note: when i18n moves to its own library, we may want to configure logging on it as well, similar to what frontend-auth does.  It probably shouldn't assume that its consumer has logging initialized.

package.json
robrap: done.

PR #167, opened 2019-05-02
fix: delay closing a form on success so a user can read the message

abutterworth:
Wait for 1000ms instead of 300ms after successfully saving before closing a form


src/profile/sagas.js
```diff
@@ -135,7 +135,7 @@ export function* handleSaveProfile(action) {
     // The preferences draft is valid if the server didn't complain, so
     // pass it through directly.
     yield put(saveProfileSuccess(accountResult, preferencesResult));
-    yield delay(300);
+    yield delay(1000);
```
albemarle: Might be nice to make this a constant and / or have it built into the confirmation component.

PR #172, opened 2019-05-10
ARCH-748: update i18n make targets

robrap:
ARCH-7481

update i18n make targets

* update extract_translations to also concat
* update push_translations to push to Transifex
* remove call to i18n.extract as part of push_translations

@albemarle: This PR adds `i18n.concat` to the `extract_translations` that the Jenkins scripts [push script](https://github.com/edx/ecommerce-scripts/blob/master/transifex/push.py#L3-L4) relies on.
I think this solves part of the problem, but having `extract_translations` do everything needed to prep source files to be committed.

Next, I plan to also add a separate script and job (called `update_sources`) to only call `extract_translations`, that can be called daily.

Let me know what you think of this PR and this plan.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
-extract_translations: ## no prerequisites so we can control order of operations
-	echo "We have to define this target due to tooling assumptions"
-	echo "Also we have to npm install using this hook b/c there's no other place for it in the current setup"
+requirements:
 	npm install
-	npm run-script i18n_extract
 
 i18n.extract:
     # Pulling display strings from .jsx files into .json files...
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,8 +1,9 @@
-extract_translations: ## no prerequisites so we can control order of operations
+requirements:
+	npm install
+
+extract_translations: | requirements i18n.extract i18n.concat
 	echo "We have to define this target due to tooling assumptions"
```
albemarle: Probably we can do without these messages now.

Makefile
```diff
@@ -19,7 +20,8 @@ tx_url1 = https://www.transifex.com/api/2/project/edx-platform/resource/frontend
 tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/frontend-app-profile/source/
 
 # push translations to Transifex, doing magic so we can include the translator comments
-push_translations: | i18n.extract
+push_translations:
```
albemarle: This still needs the precondition, I think?

PR #183, opened 2019-05-29
fix: load google fonts with a link tag

abutterworth:
The @import still exists inside edx-bootstrap, but this will make the request earlier.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,6 +4,7 @@
       <title>Learner Profile | edX</title>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -4,6 +4,7 @@
       <title>Learner Profile | edX</title>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1.0">
+      <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet">
```
douglashall: Should we do this for the other apps as well? Do you need to remove a reference to this font from somewhere else?

abutterworth: It's in edx-bootstrap. I've make a PR there which has an alternative approach: https://github.com/edx/edx-bootstrap/pull/54

PR #184, opened 2019-05-29
Move to frontend-i18n library.

douglashall:


albemarle:
Let's wait on this until Transifex jobs are up and running, just in case it's something about the move to the library that is making the Account Settings job fail.

douglashall:
@albemarle What's the downside of this causing the jobs to fail?

PR #186, opened 2019-05-30
feat: remove unused CSS with purgecss

abutterworth:
Purged CSS shrinks from 253kb for each ltr and rtl css files down to 27kb. 

Important Note: PurgeCSS will remove CSS the app actually needs if we aren't careful. This could result in a hard to identify problem in the future if a non-edX React component that has styles is used in the app. It will appear fine in development but the CSS will be stripped out in production. To fix, the package directory would need to be added to the PurgeCSS configuration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19433,6 +19433,283 @@
       "integrity": "sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==",
       "dev": true
     },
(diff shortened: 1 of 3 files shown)


davidjoy:
Do you think we need to protect ourselves against the Important Note somehow?  I worry about losing this knowledge 6 months from now when this PR comment is long gone...

albemarle:
> Do you think we need to protect ourselves against the Important Note somehow? I worry about losing this knowledge 6 months from now when this PR comment is long gone...

Could we add it to the repo's README?

abutterworth:
@albemarle Added in this PR https://github.com/edx/frontend-app-profile/pull/187

PR #194, opened 2019-06-03
feat: use postcss to handle rtl

abutterworth:
In order to work, this PR will need to upgrade to a version of frontend-i18n that sets the dir attribute on the html to ltr or rtl in all cases and does not attempt to handle stylesheet toggling.

See: https://github.com/edx/frontend-i18n/pull/12


davidjoy:
Pending reconciliation of autoprefixer and the RTL handling

PR #203, opened 2019-06-06
Update frontend-i18n version and use scripts.

davidjoy:


Makefile
```diff
@@ -1,49 +1,8 @@
...
-	# Pushing comments to Transifex...
-	./node_modules/reactifex/bash_scripts/put_comments.sh $(tx_url2)
+transifex_temp = ./temp/babel-plugin-react-intl
 
-# Pull translations from Transifex
-pull_translations:
-	tx pull -f --mode reviewed --language=$(langs)
+include node_modules/.bin/transifex-Makefile
+# to override the language list from transifex-Makefile, add a langs variable here:
+# transifex_langs = "fr,es-419"
```
albemarle: Can you hit return here?

PR #214, opened 2019-07-27
ARCH-982: temporary addition of debugging info for auth state

robrap:
### Notes: 

- Since we are tracking down a flaky issue, the sooner we can get this out the better.
- Although we are looking into problem for the new Payment MFE, the Profile MFE currently gets the most traffic, so adding this here.

### Commit Comment:

Looking into an issue where the auth state might not match.  This is
a temporary change that should ultimately be reverted once the issue
is fully involved.

This includes an upgrade to @edx/frontend-logging which does not need
to be reverted.

ARCH-982

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -34,12 +34,36 @@ const apiClient = getAuthenticatedAPIClient({
   loggingService: NewRelicLoggingService,
 });
 
+let callbackAccessToken = null;
+
+/**
+ * Temporary check of auth state vs callback access token to help debug and fix any
+ * potential problems.
+ */
+function checkAuthState(functionName, authenticationState) {
+  const userIdsMatch = !!(
+    callbackAccessToken && callbackAccessToken.user_id &&
+    authenticationState && authenticationState.authentication &&
+    authenticationState.authentication.userId &&
+    callbackAccessToken.user_id === authenticationState.authentication.userId
+  );
+  const customAttributes = {
+    functionName,
+    callbackAccessToken,
+    authenticationState,
+    userIdsMatch,
+  };
+  NewRelicLoggingService.logInfo(`checkAuthState: ${functionName}: userIdsMatch=${userIdsMatch}`, customAttributes);
+}
+
 /**
  * We need to merge the application configuration with the authentication state
  * so that we can hand it all to the redux store's initializer.
  */
 function createInitialState() {
-  return Object.assign({}, { configuration }, apiClient.getAuthenticationState());
+  const authenticationState = apiClient.getAuthenticationState();
+  checkAuthState('createInitialState', authenticationState);
+  return Object.assign({}, { configuration }, authenticationState);
 }
 
 function configure() {
@@ -66,6 +90,9 @@ function configure() {
 apiClient.ensurePublicOrAuthenticationAndCookies(
   window.location.pathname,
   (accessToken) => {
+    callbackAccessToken = accessToken;
+    checkAuthState('ensurePublicOrAuthenticationAndCookies:callback', apiClient.getAuthenticationState());
+
     const { store, history } = configure();
 
     ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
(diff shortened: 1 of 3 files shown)


src/index.jsx
```diff
@@ -34,12 +34,36 @@ const apiClient = getAuthenticatedAPIClient({
...
+}
+
 /**
  * We need to merge the application configuration with the authentication state
  * so that we can hand it all to the redux store's initializer.
  */
 function createInitialState() {
-  return Object.assign({}, { configuration }, apiClient.getAuthenticationState());
+  const authenticationState = apiClient.getAuthenticationState();
+  checkAuthState('createInitialState', authenticationState);
```
robrap: Note to reviewers: if the state doesn't match in certain cases, there is probably a quick workaround solution to use the cached callbackAccessToken instead, and we can invest more time in fixing frontend-auth later.

src/index.jsx
```diff
@@ -34,12 +34,36 @@ const apiClient = getAuthenticatedAPIClient({
...
+    authenticationState.authentication.userId &&
+    callbackAccessToken.user_id === authenticationState.authentication.userId
+  );
+  const customAttributes = {
+    functionName,
+    callbackAccessToken,
+    authenticationState,
+    userIdsMatch,
+  };
+  NewRelicLoggingService.logInfo(`checkAuthState: ${functionName}: userIdsMatch=${userIdsMatch}`, customAttributes);
```
nasthagiri: Are you outputting `functionName` twice?  Once in `customAttributes` and once after `checkAuthState:`.

src/index.jsx
```diff
@@ -34,12 +34,36 @@ const apiClient = getAuthenticatedAPIClient({
...
+  const userIdsMatch = !!(
+    callbackAccessToken && callbackAccessToken.user_id &&
+    authenticationState && authenticationState.authentication &&
+    authenticationState.authentication.userId &&
+    callbackAccessToken.user_id === authenticationState.authentication.userId
+  );
+  const customAttributes = {
+    functionName,
+    callbackAccessToken,
+    authenticationState,
```
nasthagiri: Do you need to log the entire `callbackAccessToken` and `authenticationState` object values?  Any PII concerns there?

src/index.jsx
```diff
@@ -66,13 +90,21 @@ function configure() {
...
 
     ReactDOM.render(<App store={store} history={history} />, document.getElementById('root'));
 
     if (accessToken) {
       identifyAuthenticatedUser(accessToken.userId);
     } else {
+      // TODO: Could this ever happen?
+      // - If this did happen, wouldn't we want to raise an error?
+      // - Adding error logging for now.
+      NewRelicLoggingService.logError('Empty accessToken returned from ' +
```
nasthagiri: Can you prefix these log statements so it's clear to developers from which application a log entry originated from?   Or does the `NewRelicLoggingService` already add a prefix?

src/index.jsx
```diff
@@ -34,12 +34,36 @@ const apiClient = getAuthenticatedAPIClient({
   loggingService: NewRelicLoggingService,
 });
 
+let callbackAccessToken = null;
```
nasthagiri: What are the issues with using a global variable here?

robrap: I’m not clear. Do you think there is an issue? It is scoped to the module.

src/index.jsx
robrap: Was just trying to get something done quick and didn’t know what would be helpful for debugging. We’ve done this before. I don’t think email and username to NewRelic temporarily will be an issue. If you are concerned I can edit down further.

src/index.jsx
robrap: Yes. Same with `userIdsMatch`.  The former is the message which appears on the NewRelic JS Errors page.  The second is an object that can be used in NewRelic Insight's queries.

src/index.jsx
robrap: It gets logged with a bunch of meta data in NewRelic Insights, which should include info about the transaction.  Also, I don't think anyone is looking at this particular data, because you'd have to know what you are looking for.

PR #220, opened 2019-09-06
Use frontend-base!

davidjoy:


.travis.yml
```diff
@@ -1,9 +1,5 @@
 language: node_js
-node_js:
-  - lts/*
-cache:
-  directories:
-    - "~/.npm"
+node_js: 12
 before_install:
   - npm install -g npm@latest
   - npm install -g greenkeeper-lockfile@1.14.0
```
abutterworth: Are we still using greenkeeper?

src/index.jsx
```diff
@@ -1,112 +1,31 @@
...
-      // - If this did happen, wouldn't we want to raise an error?
-      // - Adding error logging for now.
-      NewRelicLoggingService.logError('Empty accessToken returned from ' +
-        'ensurePublicOrAuthenticationAndCookies callback.');
-      identifyAnonymousUser();
-    }
-    sendPageEvent();
-  },
-);
+initialize({ messages, loggingService: NewRelicLoggingService });
```
abutterworth: I like that this is named parameters in an object vs and ordered set.

src/profile/sagas.js
```diff
@@ -1,46 +1,36 @@
...
-// Utils
-import { utils } from '../common';
-
-const { keepKeys } = utils;
-
 export function* handleFetchProfile(action) {
-  const { username } = action.payload;
-  const { authenticationUsername, userAccount } = yield select(handleFetchProfileSelector);
+  const { username, isAuthenticatedUserProfile } = action.payload;
+  const userAccount = yield select(userAccountSelector);
```
abutterworth: Could we know `isAuthenticatedUserProfile` from this `select` instead of passed in through the action?

davidjoy: We can't.  It's not calculated in the selectors anymore because some of its constituent data isn't in redux anymore (the authenticationUsername)

src/profile/components/ProfileMain.jsx
```diff
@@ -0,0 +1,37 @@
...
+import { connect } from 'react-redux';
+import { Route, Switch } from 'react-router-dom';
+
+import NotFoundPage from './NotFoundPage';
+import ConnectedProfilePage from './ProfilePage';
+
+function ProfileMain(props) {
+  const authentication = useContext(AuthenticationContext);
+
+  useEffect(() => {
```
abutterworth: Nice

.travis.yml
davidjoy: No - didn't notice that line...

src/index.jsx
davidjoy: Yeah, that was a late breaking addition. :) Much more readable.

src/profile/components/ProfilePage.jsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { StatusAlert, Hyperlink } from '@edx/paragon';
 import { connect } from 'react-redux';
-import { injectIntl, intlShape } from '@edx/frontend-i18n'; // eslint-disable-line
+
 import { sendTrackingLogEvent } from '@edx/frontend-analytics';
+import { App, AuthenticationContext } from '@edx/frontend-base';
+import { injectIntl, intlShape } from '@edx/frontend-i18n'; // eslint-disable-line
```
abutterworth: I think we can lose the `// eslint-disable-line` now. I think this was from when i18n live inside the repo as an alias.

src/profile/sagas.js
```diff
@@ -1,47 +1,37 @@
-import { push } from 'connected-react-router';
-import { all, call, delay, put, select, takeEvery } from 'redux-saga/effects';
-import { logAPIErrorResponse } from '@edx/frontend-logging';
 import { FETCH_USER_ACCOUNT_FAILURE } from '@edx/frontend-auth';
-
-// Actions
+import { App } from '@edx/frontend-base';
+import { logAPIErrorResponse } from '@edx/frontend-logging';
```
abutterworth: could we update the usage to `logApiClientError`?

PR #257, opened 2019-09-20
feat: use build scripts from frontend-builder

abutterworth:


edx-frontend-build-tools/config/webpack.dev.config.js
```diff
@@ -161,7 +126,7 @@ module.exports = Merge.smart(commonConfig, {
   // reloading.
   devServer: {
     host: '0.0.0.0',
-    port: 1995,
+    port: 8080,
```
davidjoy: This part is interesting.  Wonder if we can get the port for each app from the .env file.

edx-frontend-build-tools/config/webpack.prod.config.js
```diff
@@ -3,25 +3,25 @@
...
 const HtmlWebpackNewRelicPlugin = require('html-webpack-new-relic-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const PurgecssPlugin = require('purgecss-webpack-plugin');
 const NewRelicSourceMapPlugin = require('new-relic-source-map-webpack-plugin');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin; // eslint-disable-line prefer-destructuring
 const PostCssRtlPlugin = require('postcss-rtl');
 const PostCssAutoprefixerPlugin = require('autoprefixer');
 const CssNano = require('cssnano');
+const cwd = process.cwd();
```
davidjoy: Warning: this may be different than __dirname, right?  Does it pre-suppose that someone's running the command from a particular directory?

davidjoy: Like __dirname is guaranteed not to change, regardless of where the program was run from... I wonder if cwd() is the same

edx-frontend-build-tools/config/webpack.dev.config.js
abutterworth: Done and done!

edx-frontend-build-tools/config/webpack.prod.config.js
abutterworth: process.cwd() will be the root project path. I've updated the code a lot but that's the idea.

PR #264, opened 2019-09-24
fix: adding semantic release script

davidjoy:


package.json
```diff
@@ -59,18 +59,14 @@
     "email-prop-type": "1.1.7",
     "font-awesome": "4.7.0",
     "form-urlencoded": "3.0.0",
-    "glob": "7.1.3",
     "history": "4.7.2",
-    "i18n-iso-countries": "3.7.8",
-    "iso-countries-languages": "0.2.1",
```
robrap: This is meant to just go away?

package.json
```diff
@@ -59,18 +59,14 @@
     "email-prop-type": "1.1.7",
     "font-awesome": "4.7.0",
     "form-urlencoded": "3.0.0",
-    "glob": "7.1.3",
```
robrap: This is a dev dependency?

package.json
davidjoy: Yes.  They are unused - they got moved into frontend-i18n

package.json
davidjoy: Yup.

PR #272, opened 2019-09-25
fix: bump frontend-base to 2.2.0 to initialize segment

davidjoy:
frontend-base 2.1.x had a bug where segment wasn’t being initialized.  Now it is.

Also allowing frontend-base to use its default logging service, which is NewRelicLoggingService.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "dependencies": {
     "@edx/frontend-analytics": "3.0.0",
     "@edx/frontend-auth": "7.0.1",
-    "@edx/frontend-base": "2.1.4",
+    "@edx/frontend-base": "2.2.0",
     "@edx/frontend-component-footer": "6.0.2",
     "@edx/frontend-component-header": "1.1.2",
     "@edx/frontend-i18n": "3.0.2",
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -1,7 +1,6 @@
 import 'babel-polyfill';
 
 import { App, AppProvider, APP_ERROR, APP_READY, ErrorPage } from '@edx/frontend-base';
-import { NewRelicLoggingService } from '@edx/frontend-logging';
 import React from 'react';
 import ReactDOM from 'react-dom';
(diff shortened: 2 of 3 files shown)


src/index.jsx
```diff
@@ -35,4 +34,4 @@ App.subscribe(APP_ERROR, (error) => {
   ReactDOM.render(<ErrorPage message={error.message} />, document.getElementById('root'));
 });
 
-App.initialize({ messages: [appMessages, headerMessages], loggingService: NewRelicLoggingService });
```
robrap: Note: Your description says to initialize Segment, and all I see is logging going away. :)  Maybe you can add a comment here to better describe what is going on?  Thanks.

davidjoy: The new version of frontend-base now initializes segment - we just needed to update to it.  The latest version also allowed the loggingService parameter to be optional, so I just removed it opportunistically.  Apologies, that wasn't very clear!

package.json
```diff
@@ -43,7 +43,7 @@
   "dependencies": {
     "@edx/frontend-analytics": "3.0.0",
     "@edx/frontend-auth": "7.0.1",
-    "@edx/frontend-base": "2.1.4",
+    "@edx/frontend-base": "2.2.0",
```
davidjoy: This version bump alone fixes the segment issue.

src/index.jsx
```diff
@@ -35,4 +34,4 @@ App.subscribe(APP_ERROR, (error) => {
...
+App.initialize({ messages: [appMessages, headerMessages] });
```
davidjoy: This change is just opportunistic, as loggingService is now defaulted to NewRelicLoggingService in version 2.2.0.  Arguably should have been in a different PR.

robrap: Same PR was fine. Just wanted to ensure you wanted it.  Thanks.

robrap: Same PR was fine. Just wanted to ensure you wanted it. Thanks.

PR #291, opened 2019-10-07
Create 0002-build-time-customization-with-npm-aliases.rst

abutterworth:


docs/decisions/0002-build-time-customization-with-npm-aliases.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. code-block:: bash
+  
+  # npm install <package-name>@<type>:<branded-package>
+  
+  # npm package
+  npm install @edx/frontend-component-header@npm:@edx/frontend-component-header-edx@latest
+
+  # git repository
+  npm install @edx/frontend-component-header@git:https://github.com/edx/frontend-component-header-edx.git
```
davidjoy: Out of curiosity, do we know that using `git:` or `file:` causes node_modules to behave the same way as using a published npm package?  I feel like we've seen anecdotal evidence that using filesystem URLs causes npm to act strange... I'm somewhat concerned that we might be sending people down a confusing path.  Whacha think?

davidjoy: Don't get me wrong - I think the flexibility is important/it's desirable not to have to use npm.  But I wonder if it's... consistent.

abutterworth: Afaik git repos should work nearly identically. For `file:` I know that folders outside the project directory won't get dependencies pulled into the project's node_modules folder like normal. I think this only presents an issue if you're relying on bin scripts to be pulled into the project.

PR #294, opened 2019-10-09
Update README.rst

abutterworth:


README.rst
```diff
@@ -1,34 +1,75 @@
-|Build Status| |Codecov| |npm_version| |npm_downloads| |license| |semantic-release|
```
davidjoy: But they were pretty and our other repos have them?

README.rst
```diff
@@ -1,34 +1,75 @@
...
+
+Once the dev server is up visit http://localhost:1995/u/staff.
+
+----------
+
+
+Configuration and Deployment
+----------------------------
+
+This MFE is configured via node environment variables supplied at build time. See .env.development for the list of required environment variables. Example build syntax:
```
davidjoy: Might say to look at .env instead, since that technically corresponds to the build they're doing?  Hm.

README.rst
```diff
@@ -1,34 +1,75 @@
...
+
+.. code:: bash
+
+   NODE_ENV=development ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload' npm run build
```
davidjoy:
```suggestion
   NODE_ENV=production ACCESS_TOKEN_COOKIE_NAME='edx-jwt-cookie-header-payload' npm run build
```

And might want to point out you _need_ to supply all of the env variables, and that this is just a simple example of the syntax...

README.rst
```diff
@@ -1,34 +1,75 @@
...
+
+In this project, install requirements and start the development server
+by running:
+
+.. code:: bash
+
+   npm install
+   npm start # The server will run on port 1995
+
+Once the dev server is up visit http://localhost:1995/u/staff.
```
davidjoy: Might be worth pointing out somewhere in this doc that if you visit the _current user_'s profile you'll see an interface to edit the profile, and if you visit someone else's, you'll see the fields they've designated as public as read only.  I remember getting myself really confused about this early on and spending a lot of time trying to figure out why the page was empty. :)

davidjoy: Going along with this, maybe the initial paragraph up there could give a bit more detail on the functionality and responsibilities?

PR #326, opened 2019-12-02
build: add newrelic plugin back in as experiment

davidjoy:


package.json
```diff
@@ -88,6 +88,7 @@
     "es-check": "5.0.0",
     "glob": "7.1.3",
     "husky": "3.0.9",
+    "new-relic-source-map-webpack-plugin": "1.2.0",
```
abutterworth: No upgrade to package lock?

PR #385, opened 2020-09-21
feat: add hotjar suppression to PII

zainab-amir:
Hotjar is a behavior analytics tool that analyses website use, providing feedback through tools such as heatmaps, session recordings, and surveys. Before we enable hotjar we have to suppress all PII and sensitive information so that screen recorder doesn’t capture it. Adding a `data-hj-suppress` attribute around the sensitive data marks the field as suppressed for hotjar and shows asterisks instead of text.
Screenshot 2020-09-16 at 10 03 02 AM

For Profile MFE, PII identified and suppressed are:
- Profile Avatar
- Name
- Location
- Preferred Language
- Bio
- Education
- Username on profile
- Username in header suppressed in edx/frontend-component-header-edx#98 (PR to bump the version for profile MFE: https://github.com/edx/edx-internal/pull/3210)

Ticket: [VAN-35](https://openedx.atlassian.net/browse/VAN-35)

### DIFF
--- a/src/profile/ProfilePage.jsx
+++ b/src/profile/ProfilePage.jsx
@@ -112,9 +112,11 @@ class ProfilePage extends React.Component {
 
     return (
       <React.Fragment>
-        <h1 className="h2 mb-0 font-weight-bold">{this.props.match.params.username}</h1>
-        <DateJoined date={dateJoined} />
-        <hr className="d-none d-md-block" />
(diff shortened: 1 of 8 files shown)


waheedahmed (commented):
Don't we need to suppress the profile image and location as well?

waheedahmed (approved):
Looks good 👍

PR #397, opened 2021-01-26
Update Paragon to 13 and make the app brandable

davidjoy:
This PR is best reviewed commit-by-commit.  It includes some work to fix the linter and properly set the NODE_ENV that I wanted to fix as well since we rarely touch this MFE.  

The linting commit is a lot of changes, but they're all just formatting things for the most part.  The only one that isn't is the change to `<option value="" />` to give it an explicit `&nbsp;` of content. 

Screen Shot 2021-01-26 at 10 36 31 AM


src/profile/__snapshots__/ProfilePage.test.jsx.snap
```diff
@@ -1238,7 +1209,7 @@ exports[`<ProfilePage /> Renders correctly in various states while saving an edi
     className="profile-page-bg-banner bg-primary d-none d-md-block p-relative"
   />
   <div
-    className="container-fluid"
+    className="container"
```
abutterworth, line 1212: Are these changes to container instead of container-fluid desired?

davidjoy: Yes... the app had a max-width before, so this preserves that.  I mean, I could put it back to fluid, but it makes it super wide.  I don't feel strongly; was just trying to keep it 'the same'

src/profile/forms/Country.jsx
```diff
@@ -84,7 +84,7 @@ class Country extends React.Component {
                     value={country}
                     onChange={this.handleChange}
                   >
-                    <option value="" />
+                    <option value="">&nbsp;</option>
```
abutterworth, line 87: Intended?

davidjoy: Yeah - the linter didn't like it having some textual content.

src/profile/forms/elements/EditButton.jsx
```diff
@@ -13,7 +13,8 @@ function EditButton({
 }) {
   return (
     <Button
-      className={classNames('btn-sm btn-link', className)}
+      variant="link"
+      className={classNames('btn-sm', className)}
```
abutterworth, line 17: Button has a `size="sm"` prop: https://react-bootstrap.github.io/components/buttons/#sizes

abutterworth (approved):
A couple sanity checks, but LGTM

src/profile/forms/elements/EditButton.jsx
davidjoy: Oh, cool, I'll use that.

PR #413, opened 2021-05-03
fix: use the SITE_NAME env var in index.html

davidjoy:
This uses our SITE_NAME environment variable and removes a hard-coded “edX” in the index.html file.


sarina (approved):
Approved, but pending how we resolve Kyle's comment in the template repo

Also remember to PR into Lilac when this merges

stvstnfrd (approved):

PR #458, opened 2021-08-10
chore: update test snapshots for FontAwesome upgrade

stvstnfrd:
This fixes the failing tests on #271, our oldest PR (open since 2019!).

This changeset was (naively) generated with `npm test -- -u`

This _seems_ safe, but I admittedly don't have much experience with snapshots and don't (yet) have a _great_ sense of the possible impact (it looks like this _could_ be making FA icons larger, but I don't know if that's just how they're styled in the snapshots and whether or not they're styled differently/independently in the app) :shrug: 

So I guess this is PR is part "me trying to understand what's going on" :smile: 

This is a PR into the renovate branch. Once this lands there, that PR (#271) should automerge itself (once tests pass).


davidjoy:
IMHO if the change looks good to your eye when you compare the new with the old by loading up the UI, then this is fine.  From the snapshot alone it's impossible to tell whether it's still showing the right SVG. :)

davidjoy:
FYI, I just checked in this to the renovate PR here: https://github.com/edx/frontend-app-profile/pull/271 - I hadn't realized it was the same as this PR.  

Merging that, so I think we can close this.

davidjoy:
Apparently my merging that branch auto-closed this!  Wacky.

PR #471, opened 2021-08-30
fix: clean up unused dependencies and env variables

davidjoy:
This repo had a number of old environment variables that were no longer in use.  It also had a number of dependencies that weren't actually used in code anywhere.

The only one that was used was font-awesome, but it was only 'used' to the extent that it was imported into the stylesheet and then... not used.  They're all gone.


.env.development
```diff
@@ -12,26 +12,13 @@ LOGOUT_URL='http://localhost:18000/logout'
 MARKETING_SITE_BASE_URL='http://localhost:18000'
 ORDER_HISTORY_URL='localhost:1996/orders'
 REFRESH_ACCESS_TOKEN_ENDPOINT='http://localhost:18000/login_refresh'
-SEGMENT_KEY=null
```
binodpant, line 15: is this because 'null' is treated as actually a truthy value? we have had this issue in our other repoes too where we had to use '' vs anything else for False, and True respectively

davidjoy: Yes. null gets cast to a string and becomes truthy. Same with false. An empty string is falsy, though, when it hits JS so makes a much better default.

binodpant (approved):
lgtm since you verified none of the vars are used!

PR #472, opened 2021-08-30
fix: modernize build process, remove npm publish, use github actions

davidjoy:
This is best-reviewed commit-by-commit. 

This one PR does a few different pieces of maintenance on this repo in the interest of not dwelling too long on a low-priority MFE.  To wit: 

- It switches from travis to Github actions.
- It removes our attempt at publishing the MFE components to npm - we never used it.  (See commit for details)
- It fixes the old favicon this MFE still used.
- It fixes some incorrect environment variable defaults and an outdated doc URL.
- It removes a font we don't use anywhere else and which is just bloating the MFE.
-


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: ci
+on:
+  push:
+    branches:
+    - master
```
adamstankiewicz, line 6: Should this only be run on `master`? Wouldn't we want this to run on push to any branch?

public/index.html
```diff
@@ -4,9 +4,7 @@
       <title>Learner Profile | <%= process.env.SITE_NAME %></title>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1.0">
-      <link rel="preconnect" href="https://fonts.gstatic.com/" crossorigin>
-      <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet">
-      <link rel="shortcut icon" href="<%=webpackConfig.output.publicPath%>favicon.ico" type="image/x-icon" />
+      <link rel="shortcut icon" href="<%=htmlWebpackPlugin.options.FAVICON_URL%>" type="image/x-icon" />
```
adamstankiewicz, line 7: 🎉

.github/workflows/ci.yml
davidjoy: Hmm, well it all ran on _this_ branch... admittedly I cribbed this from frontend-app-account.

public/index.html
davidjoy: I know, right?

.github/workflows/ci.yml
davidjoy: Ah, it's the `pull_request` below that gives us what we want.  That causes it to run on PRs.

adamstankiewicz: OH, sorry. The `pull_request` was already in there?! I missed that in the initial review, my bad!

PR #491, opened 2021-10-26
build: upgrade node to v16.12.0, suggested npm is v8.1.0

binodpant:
## Why does this PR exist?

edX frontend working group has been working on upgrading node to v16 per this issue: https://github.com/openedx/frontend-wg/issues/2. This PR is a first attempt at upgrading a MFE to new node and also fix up the Github actions to use nvm to standardize on the node version checked into git.

## What do devs need to do after this PR is merged?

There is a one time node version upgrade to do locally once this PR is merged to master and synced. Devs should:

- install nvm (ideally, to manage node versions). If you do not use nvm, you can use other methods to install the node version (v16.13.0) and npm (v8.1.0) on your system. However, nvm is highly recommended because it will help you standardize on the version used by everyone else plus in Github builds. Another reason is that by simply runnign `nvm use` you can find out if you need a newer node version to match what is in your newest git repo.
- Run `nvm use` on this folder before running node/npm commands. If you don't yet have the node version, nvm will prompt you to install that version with `nvm install v16.12.0`. Do it!
- optional: remove package-lock.json (although it's not necessary)
- remove node_modules (`rm -rf node_modules/` or other OS equivalent)
- OPTION 1: if you did not remove package-lock.json: `npm ci` to install dependencies and regenerate node_modules/
- OPTION 2: if you did remove the package-lock.json: `npm install` to install deps and regenerate package-lock.json. If you do this, also please check in the newly generated package-lock.json (if there are any changes there)

## For future node/npm upgrades

Simply check in a change to the .nvmrc file to process a PR! Of course locally test it out first with the new node/npm version, but the build process should automatically pick up from the .nvmrc file

## What this PR contains
- [x] nvmrc file change to use v16 (this will make nvm use the latest available node version under v16)
- [x] github actions update to use nvmrc for the node/npm version (nvm as of now chooses v16.13.0 and v8.1.0 respectively). The `echo` line is there so that following plugin's command in github build will use the environment variable which will contain the version contained in the `.nvmrc` file, to install the correct node/npm versions using `nvm install <version>`
- [X] tested out locally (npm run build, npm run test, npm start and verify ui opens)

Screen Shot 2021-10-26 at 9 53 05 AM

package lock diff is extremely large 22k lines, due to format change in npm 7, I regenerated it from scratch:

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

Once this is synced, users will need to run `nvm use` for the first time and follow instructions to set up, if using nvm
otherwise manually install node v16 and npm v 8.1.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           node-version: "${{ env.NODE_VER }}"
       - run: make requirements
       - run: make test NPM_TESTS=build
-      - run: make test NPM_TESTS="i18n_extract is-es5 lint test"
+      - run: make test NPM_TESTS="i18n_extract lint test"
       - name: upload coverage
         uses: codecov/codecov-action@v2
         with:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/$(transi
 # This directory must match .babelrc .
 transifex_temp = ./temp/babel-plugin-react-intl
 
-NPM_TESTS=build i18n_extract lint test is-es5
+NPM_TESTS=build i18n_extract lint test
 
 .PHONY: test
 test: $(addprefix test.npm.,$(NPM_TESTS))  ## validate ci suite
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,6 @@
         "codecov": "3.8.3",
         "enzyme": "3.11.0",
         "enzyme-adapter-react-16": "1.15.6",
-        "es-check": "5.2.4",
         "glob": "7.2.0",
         "react-test-renderer": "16.14.0",
         "reactifex": "1.1.1",
@@ -5354,12 +5353,6 @@
       "integrity": "sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=",
       "dev": true
     },
-    "node_modules/ansi": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/ansi/-/ansi-0.3.1.tgz",
-      "integrity": "sha1-DELU+xcWDVqa8eSEus4cZpIsGyE=",
-      "dev": true
-    },
     "node_modules/ansi-escapes": {
       "version": "4.3.2",
       "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-4.3.2.tgz",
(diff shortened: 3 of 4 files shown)


.nvmrc
```diff
@@ -0,0 +1 @@
+v16.12.0
```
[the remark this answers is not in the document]
binodpant: done!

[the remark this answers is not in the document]
davidjoy: Yeah, I think just going with the latest 16 is fine.  Having to manually upgrade for patch/minor versions doesn't seem worthwhile in my opinion.

.github/workflows/ci.yml
```diff
@@ -8,24 +8,16 @@ on:
...
+    - name: setup node
+      uses: dcodeIO/setup-node-nvm@master
       with:
-        node-version: ${{ matrix.node-version }}
-    - run: npm install -g npm@6
+       node-version: "${{ env.NODE_VER }}"
     - run: make requirements
     - run: make test NPM_TESTS=build
-    - run: make test NPM_TESTS=${{ matrix.npm-test }}
+    - run: make test NPM_TESTS="i18n_extract is-es5 lint test"
```
adamstankiewicz, line 20: Since we've dropped support for IE 11, it should safe to remove the `is-es5` check (and the associated NPM script and `es-check` package) since our other supported browsers are [natively compatible with ES6](https://www.w3schools.com/js/js_versions.asp).

.github/workflows/ci.yml
```diff
@@ -8,24 +8,16 @@ on:
...
-        - i18n_extract
-        - is-es5
-        - lint
-        - test
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - run: echo "NODE_VER=`cat .nvmrc`" >> $GITHUB_ENV
+    - name: setup node
+      uses: dcodeIO/setup-node-nvm@master
```
adamstankiewicz, line 15: nit: might not really be an issue but does this Github Action support specifying a version so it doesn't run off of `master`?

.github/workflows/ci.yml
binodpant: ah that is indeed a good point! I just was not sure if I wanted to do it for this repo for this PR, but if @davidjoy is okay with that I can include it in there

.github/workflows/ci.yml
binodpant: good flag! I think it would be great to pin this rather than use master. And be changing versions as needed. looks like we can based on their own doc 

```
- name: Set up node using nvm
  uses: dcodeIO/setup-node-nvm@v4
```

binodpant: using v4 now

.github/workflows/ci.yml
davidjoy: Yeah, we can do that. 👍🏻  Will be a nice cleanup.

binodpant: done!

davidjoy:
Note, I've seen that 404 on @edx/brand multiple times now when installing an MFE for the first time - I think it's unrelated to this PR.

binodpant:
@davidjoy once I resolve the conflict, is this PR okay to merge? Are there any concerns about developer workflows etc? It will need a one time install of nvm to be able to use different versions (if using different for different projects).

davidjoy:
So before we merge this, we (edX folks) need to figure out how to upgrade our build pipelines as well.  We'll need to update them on an per-MFE basis as we upgrade MFEs.

Tj-Tracy:
Closing this -- its no longer needed. The Repo already supports node 16: https://github.com/openedx/frontend-app-profile/pull/530

openedx-webhooks:
Although this pull request is already closed,  I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #495, opened 2021-11-03
build: use shared browserslist config

adamstankiewicz:
* Removes custom `browserslist` configuration in favor of using a [shared configuration](https://github.com/edx/browserslist-config).
* Removes `is-es5` check in CI since ES5 was only needed for IE 11 support which is dropped. The supported browsers defined by the shared configuration all [natively support ES6](https://caniuse.com/?search=es6).

This change reduces the resultant asset bundle size.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
       matrix:
         npm-test:
         - i18n_extract
-        - is-es5
         - lint
         - test
         node: [12, 14, 16]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
       matrix:
         npm-test:
         - i18n_extract
+        - is-es6
         - lint
         - test
         node: [12, 14, 16]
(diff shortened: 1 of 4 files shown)


davidjoy:
Hmm, can/should these PRs also remove `es-check`?  Is there are reason not to?

UsamaSadiq (approved):

Jawayria:
> @Jawayria @davidjoy Let's keep `es-check`, but check for ES6 rather than ES5. Validating against ES6 will help keep some guardrails in place to let us know when we use the latest JS syntax features, but our Babel configuration doesn't transpile it down to ES6 as expected.
> 
> I think it was a bit premature to assume we'd remove it earlier when I opened these PRs 😄

@adamstankiewicz es6 checks fail on browserslist-config PRs but these tests are passing if a [new PR](https://github.com/openedx/frontend-app-account/pull/604) is created against master branch.

jsnwesson:
Browserslist-config was added in a more recent PR. https://github.com/openedx/frontend-app-profile/pull/649

PR #574, opened 2022-06-22
Paragon StatusAlert deprecation 

abdullahwaheed:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102)

### What has changed
Removed deprecated `StatusAlert` component of paragon and updated it with `Alert`

### References
[Paragon StatusAlert](https://paragon-openedx.netlify.app/components/statusalert/)
[Paragon Alert](https://paragon-openedx.netlify.app/components/alert/)


src/profile/forms/SocialLinks.jsx
```diff
@@ -158,14 +158,19 @@ class SocialLinks extends React.Component {
           ),
           editing: (
             <div role="dialog" aria-labelledby="social-links-label">
-              <form onSubmit={this.handleSubmit}>
+              <form id="editing-form" onSubmit={this.handleSubmit}>
```
davidjoy, line 161: Hmm, adding an `id` solely for a test isn't generally a good idea, since it's not actually used by the code itself and exists only to make the tests function.  I'd recommend finding the form with a `querySelector`, perhaps based on the `aria-labelledby` attribute on the form's parent.  

In general, writing tests that find elements based on accessibility attributes is actually healthy, as it helps ensure our code is annotated in meaningful ways for screen readers, etc.  Using the `aria-labelledby` here is admittedly somewhat arbitrary, but it should be unique, and helps us avoid adding otherwise unnecessary IDs.

davidjoy (approved):

PR #650, opened 2022-12-13
chore: update Paragon past version 20

cdeery:
Update paragon past v20 and any possible breaking changes
remove is-es5 dependency and update browserlist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
       matrix:
         npm-test:
         - i18n_extract
-        - is-es5
         - lint
         - test
         node: [16]
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "redux-thunk": "2.4.2",
     "regenerator-runtime": "0.13.11",
     "reselect": "4.1.7",
-    "universal-cookie": "3.1.0"
+    "universal-cookie": "4.0.4"
   },
   "devDependencies": {
     "@commitlint/cli": "17.3.0",
--- a/package.json
+++ b/package.json
@@ -25,15 +25,14 @@
     "access": "public"
   },
   "browserslist": [
-    "last 2 versions",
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -11,7 +11,6 @@ jobs:
       matrix:
         npm-test:
         - i18n_extract
-        - is-es5
         - lint
         - test
```
justinhynes, line 15: Oooh. You know, maybe this isn't the right forum... but should we be requiring any of these test suites to pass before a PR can be merged? It doesn't look like we're requiring any of these today, but should be easy to add within GitHub (in the Branch Protection rules)

justinhynes (approved):
LGTM!

.github/workflows/ci.yml
justinhynes: (asking because I thougth that we might need to remove the `is-es5` tests as required in the Branch Protection settings)

hurtstotouchfire: That is a thinker... maybe a quick parking lot?

package.json
```diff
@@ -25,15 +25,14 @@
...
   },
   "browserslist": [
-    "last 2 versions",
-    "ie 11"
+    "extends @edx/browserslist-config"
   ],
   "dependencies": {
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer": "11.6.0",
     "@edx/frontend-component-header": "3.5.0",
```
hurtstotouchfire, line 33: Are we un-pinning header / footer as part of this? Or follow up

cdeery: I wanted to unpin it as a follow-up, because its managed in a different repo and it's easier to coordinate

cdeery:
Duplicate of https://github.com/openedx/frontend-app-profile/pull/649 which has already been merged.

PR #664, opened 2023-01-11
feat: set up skills builder route

MaxFrank13:
This PR sets up an `ENABLE_SKILLS_BUILDER` environment variable that acts as a feature flag for the skills builder.

Currently, a blank `FullscreenModal` component is rendered with placeholder text:

Screen Shot 2023-01-11 at 9 15 35 AM

If the `ENABLE_SKILLS_BUILDER` variable is not present, the route will render the `NotFound` page below:

Screen Shot 2023-01-11 at 9 17 14 AM


.env.development
```diff
@@ -28,3 +28,4 @@ LEARNER_RECORD_MFE_BASE_URL='http://localhost:1990'
 COLLECT_YEAR_OF_BIRTH=true
 APP_ID=''
 MFE_CONFIG_API_URL=''
+ENABLE_SKILLS_BUILDER='true'
```
justinhynes, line 31: Should we add this to the base `.env` file as well?

Assumption: We want the base `.env` file to have a list/default of _all_ available settings to the app. The `env.development` file contains smart overrides of the base file options to facilitate development.

src/skills/SkillsBuilder.jsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <FullscreenModal
+      title="Skills Builder"
+      isOpen
+      onClose={onCloseHandle}
+      footerNode={(
+        <ActionRow>
+          <Button variant="tertiary">Go back</Button>
+          <ActionRow.Spacer />
+          <Button>Submit</Button>
```
justinhynes, line 24: Will this FullScreenModal be included in the final product? Or is this just a space saver to confirm the route is working as intended?

My only worry is that we have hardcoded english here and I want to ensure that we don't miss this later. If "Skills Builder"/"Go back"/"Submit" are all strings that will be in the final product... can we just take care of that piece now?

justinhynes: errr I checked the mocks, and these do seem to be placeholders/not final. You can disregard the above comment :)

MaxFrank13: I overlooked this! I'll work on this now and get it into this PR so it doesn't fall under the radar again

PR #676, opened 2023-01-26
Feat: display current Learning Goal in profile

jsnwesson:
### What ###
Add a Learning Goal component to Profile that displays user's current Learning Goal. 

* This PR includes a unit test that checks for a rendered Learning Goal component with mock data provided.
* The component is feature flagged and not set to be released to users until after the Skills Builder form has been finished. 
* Mock data is currently used, the integration with the back-end is for another story. 

#### Associated Jira links ####
https://2u-internal.atlassian.net/browse/APER-2244
https://2u-internal.atlassian.net/browse/APER-2181

Screenshot 2023-02-07 at 10 18 19 AM


src/profile/forms/LearningGoal.jsx
```diff
@@ -0,0 +1,92 @@
...
+      }}
+    />
+  );
+};
+
+LearningGoal.propTypes = {
+  // From Selector
+  learningGoal: PropTypes.oneOf(['advance_career', 'start_career', 'learn_something_new', 'something_else']),
+  visibilityLearningGoal: PropTypes.oneOf(['private', 'all_users']),
+  editMode: PropTypes.oneOf(['editable']),
```
MaxFrank13, line 77: Should this include 'static' as an option for `editMode`?

src/profile/forms/LearningGoal.test.jsx
```diff
@@ -0,0 +1,126 @@
...
+  config: {
+    ENVIRONMENT: 'production',
+    LANGUAGE_PREFERENCE_COOKIE_NAME: 'yum',
+  },
+  messages,
+});
+
+beforeEach(() => {
+
+});
```
MaxFrank13, line 42: Do we need this `beforeEach` for this test file?

src/profile/forms/LearningGoal.test.jsx
```diff
@@ -0,0 +1,126 @@
...
+      const learningGoalRenderer = renderer.create(
+        <LearningGoalWrapper
+          {...requiredLearningGoalProps}
+          formId="learningGoal"
+        />,
+      );
+
+      const learningGoalInstance = learningGoalRenderer.root;
+
+      expect(learningGoalInstance.findByProps({ className: 'lead' }).children).toEqual(['I want to advance my career']);
```
MaxFrank13, line 108: Just a clarifying question: what does the `findByProps({ className: 'lead' })` actually find here? Is there an element in the `<LearningGoal>` component that has that class?

MaxFrank13 (commented):
Looks great! I just had a few questions

src/profile/forms/LearningGoal.messages.jsx
```diff
@@ -0,0 +1,31 @@
...
+    id: 'profile.learningGoal.options.learn_something_new',
+    defaultMessage: 'I want to learn something new',
+    description: 'Selected by user if their goal is to learn something new.',
+  },
+  'profile.learningGoal.options.something_else': {
+    id: 'profile.learningGoal.options.something_else',
+    defaultMessage: 'Something else',
+    description: 'Selected by user if their goal is not described by the other choices.',
+  },
+});
```
justinhynes, line 29: Question around these values out of curiosity -- are we deviating from the options in the B2B Skills Builder? Were these just added as a placeholder or was there a discussion that these are the options we would launch with?

justinhynes (approved):

src/profile/forms/LearningGoal.test.jsx
jsnwesson: Nope that can be cleared out

src/profile/forms/LearningGoal.test.jsx
jsnwesson: That's right! This points specifically to the `p` element that should exist if the `learningGoal` is present.

src/profile/forms/LearningGoal.messages.jsx
jsnwesson: Ah! These are placeholders while I was just focused on the frontend. I figured I'd change these values once I started the backend integration, but I will change these to reflect the b2b version.

jsnwesson: Oh, interesting, I'm seeing that the `defaultMessage` I was referencing in the [figma design](https://www.figma.com/file/ItaWjyXC2LwPobAANvMJdr/Marketplace-Skills-Builder-MVP) is different from what exists in [APER-2199](https://2u-internal.atlassian.net/browse/APER-2199). I'm going to keep the values as they are for this story and have a bike rack set up tomorrow to confirm the intended values.

justinhynes: Sounds good. Thanks for checking up on this. I don't think it needs to necessarily block this ticket. As long as we get some clarity before MVP launch :)

.env
```diff
@@ -26,3 +26,4 @@ COLLECT_YEAR_OF_BIRTH=true
 APP_ID=''
 MFE_CONFIG_API_URL=''
 ENABLE_SKILLS_BUILDER=''
+ENABLE_LEARNING_GOAL=''
```
justinhynes, line 29: I have one additional tiny nit to suggest -- what do you think of updating the name of this feature flag to something like 'ENABLE_SKILLS_BUILDER_PROFILE' to control displaying skills-builder specific information on the profile? This could eventually include more data beyond the learning goal... so maybe we update it to be more general? 

Also, there is a "goal" field in the user profile model that isn't displayed here... so it could be potentially confusing?

jsnwesson: Totally agree on this, I'll make the change!

PR #680, opened 2023-01-30
feat: added skills context

MaxFrank13:
Adding the Context Provider for the Skills Builder. 

There is a simple example coded up to demonstrate how we would be using React Context to manipulate the state of the form.

Screen Shot 2023-01-30 at 9 00 24 AM

Screen Shot 2023-01-30 at 9 01 17 AM


src/skills-builder/data/reducer.js
```diff
@@ -0,0 +1,28 @@
+import {
+  SET_GOAL,
+  SET_JOB_TITLE,
+} from './constants';
+
+export function skillsReducer(state, action) {
```
hurtstotouchfire, line 6: I know Reducer is a react term but in the context of "Skills Builder" this is pretty funny

src/skills-builder/data/actions.js
```diff
@@ -0,0 +1,14 @@
...
+} from './constants';
+
+export const setGoal = (payload) => ({
+  type: SET_GOAL,
+  payload,
+});
+
+export const setJobTitle = (payload) => ({
+  type: SET_JOB_TITLE,
+  payload,
```
justinhynes, line 13: I assume that the storing/retrieving of Job Title was here likely for testing, but we're not planning on collecting this information as part of the MVP. 

If we are confident that the Context piece is working, can we remove the pieces that we don't need?

I'd hate to _start_ the project with potential bloat.

MaxFrank13: I included it thinking we would need to reference the value in other components or an API call, but I realize now that's not the case. This also helps get around a linting flag that requires `default export` for a file with only one export.

We _will_ need to store the jobs that the learner is interested in as those values will be used by other components. So maybe I'll just change this to be `SET_CAREER_INTERESTS` as we will actually need that action and it will get around the same linting flag.

What do you think?

justinhynes: Ah, understood. I was thinking of this from a permanent persistence angle and made a bad assumption. State != Persistence. I think I need more coffee.

With this mindset... I think I'm going to pivot my question --> Are we still missing a... state/action (not sure what the right term is, eek).

We need something to store the user's learning goal, their current job, and the future job interests. Would current job/interested job be different state/actions?

justinhynes: (follow up) -- is the scope of this ticket to define _all_ the state/actions we need for the Skills Builder?

MaxFrank13: No that is actually the scope for [this ticket](https://2u-internal.atlassian.net/browse/APER-2185).

So yes to your question but I think that's okay?

justinhynes: Ah, OK. Then my preference would be to add what is needed that meets the acceptance criteria for the current ticket and remove whatever isn't (which looks like the Job Title piece?). We can get it right when we work on the next ticket.

justinhynes (approved):

PR #681, opened 2023-01-30
fix: remove LR MFE URL building logic

justinhynes:
[APER-2240]

The Profile MFE no longer needs to understand anything about the Learner Record MFE. The Credentials IDA contains logic to determine if a request should be redirected to the Learner Record MFE so we can remove these changes from the Profile MFE.

This reverses changes originally commited as part of PR #461.


MaxFrank13 (approved):
LGTM! Thanks for cleaning this up!

PR #682, opened 2023-02-03
feat: add Skills Builder Header component, refactor to use ModalDialog

justinhynes:
[APER-2258]

This PR adds the SkillsBuilderHeader component to the profile MFE.

There was also some additional refactoring work required to display the header in the modal. The `FullscreenModal` component is not flexible enough to allow custom header designs to be displayed (only text). We have to pivot to using the `ModalDialog` family of components instead.

image

screencapture-localhost-1995-skills-2023-02-03-14_35_30


package.json
```diff
@@ -65,6 +65,7 @@
     "@edx/browserslist-config": "^1.1.1",
     "@edx/frontend-build": "12.4.19",
     "@edx/reactifex": "2.1.1",
+    "@testing-library/react": "11.2.7",
```
justinhynes, line 68: This is a commonly used package in our frontend projects that, in my opinion, make writing test a bit easier... so I added it along with my other changes.

MaxFrank13: This is something I discussed with @jsnwesson. In some of the other MFEs, everything from `@testing-library/react` gets exported by the `setupTest.js` file. The Profile MFE doesn't have much in that file yet so I think it's _fine_ to import here. Although, when the SkillsBuilder eventually gets it's own MFE, we'll probably end up moving any imports from `@testing-library/react` to that `setupTest.js` file

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+const SkillsBuilderHeader = () => {
+  const { formatMessage } = useIntl();
+
+  return (
+    <div className="d-flex">
+      <img src={edXLogo} alt="edx-logo" className="mt-2 h-50" />
+      <div className="ml-5 vertical-line" />
+      <div className="w-100 ml-5">
+        <p className="h1 header-title">
```
MaxFrank13, line 14: This can actually be an `<h1>` and the subheading can be an `<h2>` — mostly for accessibility

src/skills-builder/skills-builder-header/skillsBuilderHeader.scss
```diff
@@ -0,0 +1,8 @@
+.vertical-line {
+  border-left: 7px solid #D23228;
+  transform: rotate(13deg);
+}
+
+.header-title {
```
MaxFrank13, line 6: Instead of creating a CSS class for the `header-title`, you can instead use the [color utility classes from Paragon](https://paragon-openedx.netlify.app/foundations/colors).

So for this one, I think it would be `text-warning-300`. The hex values from Figma _should_ all be represented somewhere in the Paragon spec

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
```diff
@@ -0,0 +1,9 @@
+.skills-builder-modal {
+  .pgn__modal-close-button{
+    color: #ffffff !important;
+  }
+
+  .pgn__modal-hero-bg {
+    background-color: #00262B;
+  }
```
MaxFrank13, line 8: We generally don't want to modify the classes prefixed with `pgn__`. I'd recommend using utility classes for these instead, which would also eliminate the need for the SCSS file

MaxFrank13 (changes requested):
Looks great! I just had a few requests

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
justinhynes: Hey Max -- I think this advice goes against most of the advice I've had here in the past -- which is to use _styling_ like headers, but not to use header (h1/h2) components themselves.

Dumb question -- Since this is an actual page header/hero, is this why we would use an H1 here? (I'm also under the impression that there can only be a single h1 on a page for accessibility concerts)

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
justinhynes: Two things here:
-- I needed to override the color of the close button, which is part of the Paragon component. This is how it was done in the B2B version. How would I override the color of the button otherwise?
-- The background color of the hero doesn't exist as a utility class but this is the color used in the Figma mocks. What do I do here then?

src/skills-builder/skills-builder-header/skillsBuilderHeader.scss
justinhynes: This is the color in the Figma mocks but it doesn't have a utility class in Paragon. I had added this to match the mocks.

package.json
justinhynes: I had noticed this as well but figured this MFE is behind on what may be considered "standard configuration" these days. The benefits of having the test methods outweighed some of the extra work to bring it in line with more "modern" MFEs though.

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
MaxFrank13: From what I've gathered, we dp want to use them because, without any hX elements, we have no way of telling the screen reader what the hierarchy of the content is. I wrote [this doc](https://docs.google.com/document/d/1T5Jh8Gv1kPRuwcxY82bmdCZNxTJaFoMF-RbuEwHoIyk/edit?usp=sharing) as prep for a Frontend Study Group about headings with some help from Jeff and it may clear things up.

Given that this is the Skills Builder page, I think "Skills Builder" should be the h1, or h2 at the least. And on second thought, perhaps the subtitle is fine the way you have it — it doesn't represent a section of content and therefore, doesn't require a heading element.

src/skills-builder/skills-builder-header/skillsBuilderHeader.scss
MaxFrank13: I believe this hex color does exist in Paragon as `warning-300`. Are you looking at the Open edX theme perhaps?

Screen Shot 2023-02-06 at 11 59 39 AM

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
MaxFrank13: 1. Good point! It appears we don't have a prop in the `ModalDialog` for modifying the color of the icon. You can definitely do it this way and it won't cause any issues. I think it's more of a nit at this point, but I would go for something like:
```
.skills-builder-modal {
  button[aria-label="Close"][type="button"]{
    color: #ffffff;
  }
}
```

I personally am not a huge fan of using the `!important` tag unless we absolutely have to. The above method gets around it by having higher specificity than the current styles.

2. You can add a `className` to the `ModalDialog.Hero.Background` component and give it a utility class `bg-primary-500` and that should take care of that without having to do any custom SCSS.

FWIW, this is mostly a product of a discussion that has gone on in Frontend groups. The consensus is that our design system _should_ be able to do all of this without any custom styling. So the fact that we need to do this is a sign that perhaps the system has room for improvement in this area

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
justinhynes: OK. I'll go ahead and make this change. Thanks for the clarification. I'll make the main heading an `H1` and the subheading an `H2` -- this makes the most sense to me.

src/skills-builder/skills-builder-header/skillsBuilderHeader.scss
justinhynes: I _was_ looking at the Open edX theme. I had no idea until today that you could change the theme on the Paragon site :-O. Thanks for this! I'll fix this up.

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
justinhynes: 1. Your suggestion for the first point worked great. Thanks for the help here, even though I do find this harder to understand than the above solution (to me). 
2. This is super straight forward too. Thanks.

MaxFrank13 (approved):
LGTM! Thanks for making the changes!

src/skills-builder/skills-builder-header/messages.js
```diff
@@ -0,0 +1,16 @@
...
+
+const messages = defineMessages({
+  skillsBuilderHeaderTitle: {
+    id: 'skills.builder.header.title',
+    defaultMessage: 'Skills Builder',
+    description: 'Title for the Skills Builder feature',
+  },
+  skillsBuilderHeaderSubheading: {
+    id: 'skills.builder.header.subheading',
+    defaultMessage: 'Let edX be your guide',
```
sarina, line 11: Hi - this pull request (specifically this line) introduced a bug:

https://github.com/openedx/frontend-app-profile/issues/686

Could you please address by refactoring the edX-specific language out of strings? Thank you.

PR #690, opened 2023-02-14
feat: Add Algolia support to Profile MFE for Skills Builder

justinhynes:
[APER-2261]

This PR adds Algolia support to the Profile MFE and the upcoming Skills Builder feature.

* Adds new dependency for the `algoliasearch` package
* Add new config to support Algolia
* Update MFE configuration so we can access the new configuration variables
* Add hook to initialize Algolia client and return job and product Algolia indexes (based on settings)
* Update SkillsBuilderModal to add test code that displays the results of querying Algolia


MaxFrank13 (approved):
Looks great! Thanks for making the changes to accommodate [APER-2185](https://github.com/openedx/frontend-app-profile/pull/685)

PR #693, opened 2023-02-21
APER-2186: adding goal and job title selection

MaxFrank13:
### Description

This PR adds a dropdown selection for the learner's current goal as well as an `Autosuggest` component for the learner to select their current job.

The list of options in the `Autosuggest` is retrieved from Algolia. Whenever the learner provides input, the component will query Algolia and update the list. This component is built to be reusable for other parts of the `SkillsBuilder`.

When a learner selects a job title, this will enable them to proceed in the `SkillsBuilder`. There are also checkboxes available for learners who are currently students or looking for work. Selecting either of these will also enable the learner to proceed.

### JIRA
[APER-2186](https://2u-internal.atlassian.net/browse/APER-2186)

#### Default view:
Screenshot 2023-02-21 at 11 40 45 AM

#### Goal has been selected:
Screenshot 2023-02-21 at 11 42 12 AM

#### Job title selected OR checkbox is checked:
Screenshot 2023-02-21 at 11 42 52 AM


.env.development
```diff
@@ -28,7 +28,7 @@ APP_ID=''
...
 ENABLE_SKILLS_BUILDER='true'
 ENABLE_SKILLS_BUILDER_PROFILE=''
-ALGOLIA_APP_ID=''
-ALGOLIA_JOBS_INDEX_NAME=''
-ALGOLIA_PRODUCT_INDEX_NAME=''
-ALGOLIA_SEARCH_API_KEY=''
+ALGOLIA_APP_ID='testing2MLYZ9WWU0'
+ALGOLIA_JOBS_INDEX_NAME='stage_taxonomy'
+ALGOLIA_PRODUCT_INDEX_NAME='product'
+ALGOLIA_SEARCH_API_KEY='b0004be3c12168d98a6d6b793fcd46b9'
```
justinhynes, line 34: Please don't commit these values to source control.

.env.test
```diff
@@ -23,7 +23,7 @@ LEARNER_RECORD_MFE_BASE_URL='http://localhost:1990'
...
 APP_ID=''
 MFE_CONFIG_API_URL=''
-ALGOLIA_APP_ID=''
-ALGOLIA_JOBS_INDEX_NAME=''
-ALGOLIA_PRODUCT_INDEX_NAME=''
-ALGOLIA_SEARCH_API_KEY=''
+ALGOLIA_APP_ID='testing2MLYZ9WWU0'
+ALGOLIA_JOBS_INDEX_NAME='stage_taxonomy'
+ALGOLIA_PRODUCT_INDEX_NAME='product'
+ALGOLIA_SEARCH_API_KEY='b0004be3c12168d98a6d6b793fcd46b9'
```
justinhynes, line 29: Please don't commit these values to source control.

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -48,8 +48,10 @@ const SkillsBuilderModal = () => {
           </ModalDialog.Hero.Content>
         </ModalDialog.Hero>
 
+        <Stepper.Header />
+
         <ModalDialog.Body>
-          <Container size="lg">
+          <Container size="md">
             <Stepper.Step eventKey={STEP1} title="Select preferences">
```
justinhynes, line 55: This caught my eye but I know it's not part of this PR -- is the `title` displayed anywhere? Should this suport i18n/translation?

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const CareerInterestSelect = () => (
+  <div>
+    <h4>What careers are you interested in?</h4>
```
justinhynes, line 5: Should this be hard-coded English?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
```diff
@@ -0,0 +1,60 @@
...
+import PropTypes from 'prop-types';
+import {
+  Form,
+} from '@edx/paragon';
+import { logError } from '@edx/frontend-platform/logging';
+import { useAlgoliaSearch } from '../../utils/search';
+
+const JobTitleAutosuggest = ({ onSelected }) => {
+  // We only need to destructure the job search index here
+  const [,, jobSearchIndex] = useAlgoliaSearch();
```
justinhynes, line 11: I assumed we would be passing around the index where needed instead of re-instantiating the objects with an additional call to useAlgoliaSearch. Do we have access to to passing in the index here to this component?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const [jobInput, setJobInput] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [jobList, setJobList] = useState([]);
+
+  const queryForJobs = (query = '') => {
+    jobSearchIndex.search(query).then(({ hits }) => {
+      setJobList(hits);
+      setIsLoading(false);
```
justinhynes, line 20: Is the utility function from `search.jsx` not good enough here? This seems like some redundant code. Maybe I don't understand the use case here.

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
```diff
@@ -0,0 +1,60 @@
...
+      ))}
+    </Form.Autosuggest>
+  );
+};
+
+JobTitleAutosuggest.propTypes = {
+  onSelected: PropTypes.func.isRequired,
+};
+
+export default JobTitleAutosuggest;
```
justinhynes, line 60: Instead of making our own component for this... would this be a good place to use the <InstantSearch /> package and component that B2B uses? 

We would already have access to the client/index we would pass into the object and then it's less code for us to maintain.

src/skills-builder/skills-builder-modal/select-preferences/messages.js
```diff
@@ -0,0 +1,61 @@
...
+    defaultMessage: 'First, tell us what you want to achieve',
+    description: 'Prompts the user to select their current goal.',
+  },
+  selectLearningGoal: {
+    id: 'select.learning.goal',
+    defaultMessage: 'Select a goal',
+    description: 'Placeholder text for the goal selection component.',
+  },
+  learningGoalStartCareer: {
+    id: 'learning.goal.start.career',
```
justinhynes, line 20: Nit: Can we make the ids consistent here?  We have `learning.goal.start.career` here, but then next is `learning.goal.advance_career`. 

Suggestion... we pick one naming convention and stick to it. I'm partial to `name_with_underscores` as the final value...
```
learning.goal.start_career
learning.goal.advance_career
learning.goal.something_new
learning.goal.something_else
``` etc.

File: src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx (hunk 3 above)
Comment (MaxFrank13) [reply src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx:55]: Yes it does. Nice catch! The title attribute is the same text used to display in the `Stepper.Header`

File: src/skills-builder/test/SkillsBuilder.test.jsx [hunk 9]
```diff
@@ -1,19 +1,73 @@
...
 );
 
 describe('skills-builder', () => {
-  it('should render a Skills Builder modal', () => {
+  beforeAll(async () => {
+    await mergeConfig({
+      ALGOLIA_APP_ID: process.env.ALGOLIA_APP_ID || null,
+      ALGOLIA_JOBS_INDEX_NAME: process.env.ALGOLIA_JOBS_INDEX_NAME || null,
+      ALGOLIA_PRODUCT_INDEX_NAME: process.env.ALGOLIA_PRODUCT_INDEX_NAME || null,
+      ALGOLIA_SEARCH_API_KEY: process.env.ALGOLIA_SEARCH_API_KEY || null,
```
justinhynes, line 36: I don't think we should be trying to use a live service in our tests here.

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
MaxFrank13: This was just placeholder for the next ticket, APER-2249, but yes it will need to be translated so I'll set that up here so it's accounted for

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
MaxFrank13: Yes we can do this

justinhynes (changes requested):
Let me know if you want to chat more about my comments.

src/skills-builder/test/SkillsBuilder.test.jsx
justinhynes: Also, if these values are null, like they would be for most people... would these tests fail?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
MaxFrank13: From what I could tell, the utility functions don't return jobs based on a search query. There is one that returns job information but it requires job names. The query in this PR returns a list of job names to choose from based on whatever string you provide it. Perhaps I missed something?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
MaxFrank13: Currently working on this and believe it may provide a solution to some of the other concerns as well 👍

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const handleAutoSuggestChange = (title) => {
+    setIsLoading(true);
+    setJobInput(title);
+    queryForJobs(title);
+  };
+
+  useEffect(() => {
+    queryForJobs();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
MaxFrank13: This would actually return an array of the first 20 jobs. I'm not sure how it determines which 20 to show, however. It didn't have a ton of use, though. It basically just renders a list of jobs without the user needing to type anything in.

After speaking with Justin yesterday, we found that perhaps using the `InstantSearch` component from the `react-instantsearch-hooks-web` package would be a better option than the implementation here. Still working on updating the PR with those changes as testing is proving somewhat difficult.

src/skills-builder/skills-builder-modal/select-preferences/JobTitleAutosuggest.jsx
MaxFrank13: The `useAlgoliaSearch` function is now called inside the `SkillsContextProvider` so that this data is available to all components in the `SkillsBuilder`.

src/skills-builder/test/SkillsBuilder.test.jsx
MaxFrank13: These values have been removed except for `ALGOLIA_JOBS_INDEX_NAME` which has been replaced with a default `test-job-title-index` string

src/skills-builder/skills-builder-modal/SelectPreferences.jsx
```diff
@@ -1,63 +0,0 @@
-import React, { useContext } from 'react';
```
MaxFrank13, line 1: This was a duplicate file that I missed earlier. The REAL `SelectPreferences` component lives under the `select-preferences` directory

src/skills-builder/skills-builder-context/SkillsBuilderProvider.jsx
```diff
@@ -1,12 +1,24 @@
...
+
+  const value = useMemo(() => ({
+    state,
+    dispatch,
+    algolia: {
+      searchClient,
+      productSearchIndex,
+      jobSearchIndex,
+    },
+  }), [state, searchClient, productSearchIndex, jobSearchIndex]);
```
MaxFrank13, line 21: We will need to reach for `searchClient` at least one more time in the app. Storing it in context gives us easy access to it; rather than passing it down through various parent components.

justinhynes (approved):
Thanks for making all those changes! LGTM

PR #702, opened 2023-02-28
feat: SkillsBuilder career interests selection

MaxFrank13:
### Description

This PR adds a selection component that enables the user to select potential career interests from a list of job titles. The `JobTitleInstantSearch` component was able to be reused here.

In addition to writing tests for new components, some other tests in the `SkillsBuilder.test.jsx` suite were modified to be a little more robust. 💪  There was also a spelling typo for the `REMOVE_CAREER_INTEREST` action that was fixed.

There is one custom component here: the `CareerInterestCard`. It is a simple card that employs utility classes from Paragon and will dismiss itself when it's corresponding close button is selected.

#### `JobTitleInstantSearch` with no selections
Screenshot 2023-02-28 at 12 26 24 PM

#### Component with three selections
Screenshot 2023-02-28 at 12 27 12 PM

#### Result of clicking remove icon
Screenshot 2023-02-28 at 12 28 51 PM

### DIFF
--- a/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
+++ b/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
@@ -1,0 +1,35 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import {
+  IconButton, Icon,
+} from '@edx/paragon';
+import { useIntl } from '@edx/frontend-platform/i18n';
+import { Close } from '@edx/paragon/icons';
+import { SkillsBuilderContext } from '../../skills-builder-context';
+import { removeCareerInterest } from '../../data/actions';
+import messages from './messages';
+
+const CareerInterestCard = ({ interest }) => {
+  const intl = useIntl();
+  const { dispatch } = useContext(SkillsBuilderContext);
+
+  return (
+    <div className="d-flex justify-content-between align-items-center pb-2 pr-2 pl-4 rounded shadow-sm">
+      <p className="pt-4">
+        {interest}
+      </p>
+      <IconButton
+        iconAs={Icon}
+        src={Close}
+        alt={`${intl.formatMessage(messages.removeCareerInterestButtonAltText)} ${interest}`}
+        onClick={() => dispatch(removeCareerInterest(interest))}
+      />
+    </div>
+  );
+};
+
+CareerInterestCard.propTypes = {
+  interest: PropTypes.string.isRequired,
+};
+
+export default CareerInterestCard;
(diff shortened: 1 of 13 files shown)


src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <div className="d-flex justify-content-between align-items-center pb-2 pr-2 pl-4 rounded shadow-sm">
+      <p className="pt-4">
+        {interest}
+      </p>
+      <IconButton
+        iconAs={Icon}
+        src={Close}
+        alt={`Remove ${interest}`}
```
justinhynes, line 21: The `alt` here refers to alternative text, right? "Remove" here would be hardcoded English, right? Is there a way to handle this with our i18n capabilities?

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
```diff
@@ -1,16 +1,44 @@
...
-    </p>
-  </div>
-);
+const CareerInterestSelect = () => {
+  const { state, dispatch, algolia } = useContext(SkillsBuilderContext);
+  const { careerInterests } = state;
+  const { searchClient } = algolia;
+
+  const handleCareerInterestSelect = (value) => {
+  // 'onSelected' function was adding a null value to the 'careerInterests' array during the onChange event
```
justinhynes, line 18: Nit: I'm a little confused by the comment. I'm not sure what it's trying to tell me.

MaxFrank13: That's a fair question — I don't think I made this very clear in the comment. It's referring to the `if` statement below that checks for a `value` to exist (as well as checking the length of the `careerInterests` array). Without checking for the value to exist, the code will dispatch a `null` value to add to the `careerInterests` array every time a user types input. So we end up with empty cards.

`onSelected` refers to the attribute of the `Form.Autosuggest` component from Paragon. I was under the impression that the code only fired when a selection was made. But it appears to fire after any `onChange` event as well. This perplexed me a bit because `Form.Autosuggest` also has a separate `onChange` attribute that we are using already.

justinhynes: Ah. Gotcha. That's pretty good context. Can we try and condense the context and improve this comment?

justinhynes (approved):
A couple of minor items but overall looks good!

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
MaxFrank13: Yes absolutely 👍

PR #706, opened 2023-03-06
APER-2187 Render jobs and related skills for Skills Builder

MaxFrank13:
### Description
This PR adds a set of `SelectableBox` components to Step 2 of the `SkillsBuilder`. Each box contains a job title with related skills. To retrieve the related skills, the learner's `careerInterests` are passed to Algolia using a utility function. The skills are then used to retrieve course recommendations using another Algolia utility function.

### JIRA
[APER-2187](https://2u-internal.atlassian.net/browse/APER-2187) — Render jobs and related skills

### Screenshot
Screenshot 2023-03-06 at 11 48 36 AM

### DIFF
--- a/src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
+++ b/src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
@@ -2,7 +2,7 @@ import React, { useState, useContext } from 'react';
 import {
   Button, Container, Stepper, ModalDialog,
 } from '@edx/paragon';
-import { FormattedMessage, useIntl } from '@edx/frontend-platform/i18n';
+import { useIntl } from '@edx/frontend-platform/i18n';
 import { useHistory } from 'react-router';
 import {
   STEP1, STEP2,
@@ -17,7 +17,7 @@ import ViewResults from './view-results/ViewResults';
 import headerImage from '../images/headerImage.png';
 
 const SkillsBuilderModal = () => {
-  const intl = useIntl();
+  const { formatMessage } = useIntl();
   const { state } = useContext(SkillsBuilderContext);
   const { careerInterests } = state;
   const [currentStep, setCurrentStep] = useState(STEP1);
@@ -49,12 +49,12 @@ const SkillsBuilderModal = () => {
         <Stepper.Header />
 
         <ModalDialog.Body>
-          <Container size="md">
-            <Stepper.Step eventKey={STEP1} title={intl.formatMessage(messages.selectPreferences)}>
+          <Container size="md" className="p-4.5">
+            <Stepper.Step eventKey={STEP1} title={formatMessage(messages.selectPreferences)}>
               <SelectPreferences />
             </Stepper.Step>
 
-            <Stepper.Step eventKey={STEP2} title={intl.formatMessage(messages.reviewResults)}>
+            <Stepper.Step eventKey={STEP2} title={formatMessage(messages.reviewResults)}>
               <ViewResults />
             </Stepper.Step>
           </Container>
@@ -63,14 +63,14 @@ const SkillsBuilderModal = () => {
         <ModalDialog.Footer>
           <Stepper.ActionRow eventKey={STEP1}>
             <Button variant="outline-primary" onClick={onCloseHandle}>
-              <FormattedMessage {...messages.goBackButton} />
+              {formatMessage(messages.goBackButton)}
             </Button>
             <Stepper.ActionRow.Spacer />
             <Button
               onClick={() => setCurrentStep(STEP2)}
               disabled={careerInterests.length === 0}
             >
-              <FormattedMessage {...messages.nextStepButton} />
+              {formatMessage(messages.nextStepButton)}
             </Button>
           </Stepper.ActionRow>
           <Stepper.ActionRow eventKey={STEP2}>
@@ -78,11 +78,11 @@ const SkillsBuilderModal = () => {
               variant="outline-primary"
               onClick={() => setCurrentStep(STEP1)}
             >
-              <FormattedMessage {...messages.goBackButton} />
+              {formatMessage(messages.goBackButton)}
             </Button>
             <Stepper.ActionRow.Spacer />
             <Button onClick={onCloseHandle}>
-              <FormattedMessage {...messages.exitButton} />
+              {formatMessage(messages.exitButton)}
             </Button>
           </Stepper.ActionRow>
         </ModalDialog.Footer>
--- a/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
+++ b/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import { getConfig } from '@edx/frontend-platform';
-import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import { useIntl } from '@edx/frontend-platform/i18n';
 import { Stack, Row, Col } from '@edx/paragon';
 import { InstantSearch } from 'react-instantsearch-hooks-web';
 import JobTitleInstantSearch from './JobTitleInstantSearch';
@@ -10,6 +10,7 @@ import { SkillsBuilderContext } from '../../skills-builder-context';
 import messages from './messages';
 
 const CareerInterestSelect = () => {
+  const { formatMessage } = useIntl();
   const { state, dispatch, algolia } = useContext(SkillsBuilderContext);
   const { careerInterests } = state;
   const { searchClient } = algolia;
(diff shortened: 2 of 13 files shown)


src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -17,7 +17,7 @@ import ViewResults from './view-results/ViewResults';
 import headerImage from '../images/headerImage.png';
 
 const SkillsBuilderModal = () => {
-  const intl = useIntl();
+  const { formatMessage } = useIntl();
```
MaxFrank13, line 20: I refactored the way the `formatMessage` is used in several components. I felt it was better to have it uniform throughout the app.

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
```diff
@@ -26,10 +27,13 @@ const CareerInterestSelect = () => {
...
     <Stack gap={2}>
       <h4>
-        <FormattedMessage {...messages.careerInterestPrompt} />
+        {formatMessage(messages.careerInterestPrompt)}
       </h4>
       <InstantSearch searchClient={searchClient} indexName={getConfig().ALGOLIA_JOBS_INDEX_NAME}>
-        <JobTitleInstantSearch onSelected={handleCareerInterestSelect} />
+        <JobTitleInstantSearch
+          onSelected={handleCareerInterestSelect}
+          placeholder={formatMessage(messages.careerInterestInputPlaceholder)}
```
justinhynes, line 35: What is the purpose of passing in a static placeholder string as a prop to this component? Wouldn't this just be rendered as part of the component?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
```diff
@@ -28,17 +29,20 @@ const JobTitleSelect = () => {
...
+        />
       </InstantSearch>
       <Form.Group>
         <Form.CheckboxSet
           name="other-occupations"
           onChange={handleCheckboxChange}
         >
-          <Form.Checkbox value="student">
+          <Form.Checkbox value="Student">
             <FormattedMessage {...messages.studentCheckboxPrompt} />
```
justinhynes, line 43: Should we have updated this line (and line 46) with the rest of the i18n changes?

src/skills-builder/skills-builder-modal/select-preferences/SelectPreferences.jsx
```diff
@@ -14,8 +14,8 @@ const SelectPreferences = () => {
   const { currentGoal, currentJobTitle } = state;
 
   return (
-    <>
-      <p className="lead mb-5">
+    <Stack gap={4}>
+      <p className="lead">
         <FormattedMessage {...messages.skillsBuilderDescription} />
```
justinhynes, line 19: Is this another line where we should have updated the way we're handling i18n?

MaxFrank13: It looks like I missed a few more of these. I'll run through the app and make sure they're all the same

src/skills-builder/skills-builder-modal/view-results/RelatedSkillsSelectableBoxSet.jsx
```diff
@@ -0,0 +1,51 @@
...
+          key={job.id}
+          type="radio"
+          value={job.name}
+          aria-label={job.name}
+          inputHidden={false}
+        >
+          <p>{job.name}</p>
+          <Stack gap={2} className="align-items-start">
+            <p className="heading-label x-small">{formatMessage(messages.relatedSkillsHeading)}</p>
+            {job.skills.slice(0, 5).map((skill, idx) => (
```
justinhynes, line 32: Are we just picking/displaying the first 5 skills? Are these selected based on their signifigance? Or some other ranking system?

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
MaxFrank13: The `JobTitleInstantSearch` component is used in two places so we needed somewhere to pass a unique placeholder for each instance

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -1,7 +1,81 @@
...
+  useEffect(() => {
+    const getJobs = async () => {
+      // fetch list of jobs with related skills
+      const jobInfo = await searchJobs(jobSearchIndex, careerInterests);
+
+      // fetch course recommendations based on related skills for each job
+      const results = await Promise.all(jobInfo.map(async (job) => {
+        const formattedSkills = job.skills.map(skill => skill.name);
+
+        const recommendations = await getProductRecommendations(productSearchIndex, 'course', formattedSkills);
```
justinhynes, line 32: IIRC, in the B2B version, after retrieving the course recommendations, they have some logic to de-dupe courses the learner is already enrolled in. Will we be covering this somewhere for course recommendations?

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -1,7 +1,81 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const getJobs = async () => {
+      // fetch list of jobs with related skills
+      const jobInfo = await searchJobs(jobSearchIndex, careerInterests);
+
+      // fetch course recommendations based on related skills for each job
+      const results = await Promise.all(jobInfo.map(async (job) => {
+        const formattedSkills = job.skills.map(skill => skill.name);
```
justinhynes, line 30: Nit: Indentation/spacing looks weird here?

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
justinhynes: Ahhhhhh, gotcha. Thanks.

src/skills-builder/skills-builder-modal/view-results/RelatedSkillsSelectableBoxSet.jsx
MaxFrank13: Good catch. This is just taking the top five that are returned, no sorting done. We can use an array method to sort by significance, which I think is probably best?

justinhynes: Sounds decent to me! I think this would certainly be a great thing to have for MVP.

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
MaxFrank13: Another good catch! This will be covered in the next ticket. It looks like they are [using an API in the LMS](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/master/src/components/skills-quiz/data/service.js#L28-L32) to retrieve the information on enrollments. As far as I can tell, we should be able to do something pretty similar?

justinhynes: Yep, absolutely! Thanks, Max.

src/skills-builder/skills-builder-modal/view-results/RelatedSkillsSelectableBoxSet.jsx
```diff
@@ -0,0 +1,57 @@
...
+  const { formatMessage } = useIntl();
+  const isExtraSmall = useMediaQuery({ maxWidth: breakpoints.extraSmall.maxWidth });
+
+  const renderTopFiveSkills = (skills) => {
+    const topFiveSkills = skills.sort((a, b) => b.significance - a.significance).slice(0, 5);
+    return (
+      topFiveSkills.map(skill => (
+        <Chip key={skill.id}>
+          {skill.name}
+        </Chip>
```
justinhynes, line 19: Would it be difficult to add a test for this component/functionality?

justinhynes: Should we be coding more protectively around this functionality? Do we need to ensure there are more than 5 entries? Is it possible that we run into any errors if we try to slice an array with < 5 items in it?

justinhynes: (I'm messing around in a JSFiddle sandbox and it looks like maybe this won't throw an exception if we try to slice() beyond the size of the array... so maybe this isn't as much of an issue as I thought it would be)

MaxFrank13: Yes, I don't think the `slice` will cause any issues for us here.

In regards to the first question, it would be nice to have a test to check the order of skills, ensuring that the ones with highest significance are placed at the top

src/skills-builder/test/SkillsBuilder.test.jsx
```diff
@@ -149,4 +165,29 @@ describe('skills-builder', () => {
...
+        },
+      ),
+    );
+    await act(async () => {
+      fireEvent.click(screen.getByRole('button', { name: 'Next Step' }));
+    });
+
+    const chipComponents = document.querySelectorAll('.pgn__chip');
+    expect(chipComponents[0].textContent).toEqual('finding shiny things');
+    expect(chipComponents[1].textContent).toEqual('mining');
```
justinhynes, line 189: Thanks for improving this test to include checking that the skills/significance is working as expected.

A nit: Next time, I think this would be better if we have a test explicitly for the component the functionality belongs to. Just as we want to beware functions that "do too much", we also want to beware tests that "do too much". It is also hard to tell what our test coverage is for our subcomponents when do it this way. 

This is better than no tests for the functionality at all though.

justinhynes (approved):

PR #725, opened 2023-03-28
Revert "feat: course recommendations for Skills Builder"

MaxFrank13:
Upon merging the original PR, our stage build for the Profile MFE began failing in GoCD. While we do not have any reason to believe these changes caused the issue, we are going to see if reverting has any effect, or perhaps at least provides another data point for us to work from.


justinhynes (approved):
Thanks for trying this out (sorry that this is turning into a pain in the rear).

PR #739, opened 2023-04-28
APER-2333: Dispatch segment events from JS components

MaxFrank13:
This adds analytics events to track a user's progress through the Skills Builder. These events are outlined in more detail on the [Events to monitor KPIs for Skills Builder](https://2u-internal.atlassian.net/wiki/spaces/microb/pages/395477339/Eventing+to+monitor+KPIs+for+Skills+Builder) page.

Several event handle functions have been modified/created and tests written to check their behavior.

### DIFF
--- a/src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
+++ b/src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
@@ -1,8 +1,9 @@
 import React, { useState, useContext } from 'react';
 import {
-  Button, Container, Stepper, ModalDialog,
+  Button, Container, Stepper, ModalDialog, Form,
 } from '@edx/paragon';
 import { useIntl } from '@edx/frontend-platform/i18n';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { useHistory } from 'react-router';
 import {
   STEP1, STEP2,
@@ -19,13 +20,41 @@ import headerImage from '../images/headerImage.png';
 const SkillsBuilderModal = () => {
   const { formatMessage } = useIntl();
   const { state } = useContext(SkillsBuilderContext);
-  const { careerInterests } = state;
+  const { currentGoal, currentJobTitle, careerInterests } = state;
   const [currentStep, setCurrentStep] = useState(STEP1);
 
   const history = useHistory();
 
   const onCloseHandle = () => {
     history.goBack();
+  };
+
+  const sendActionButtonEvent = (eventSuffix) => {
+    sendTrackEvent(
+      `edx.skills_builder.${eventSuffix}`,
+      {
+        app_name: 'skills_builder',
+        category: 'skills_builder',
+        learner_data: {
+          current_goal: currentGoal,
+          current_job_title: currentJobTitle,
+          career_interests: careerInterests,
+        },
+      },
+    );
+  };
+
+  const nextStepHandle = () => {
+    setCurrentStep(STEP2);
+    sendActionButtonEvent('next_step');
+  };
+  const exitButtonHandle = () => {
+    sendActionButtonEvent('exit');
+    onCloseHandle();
+  };
+  const closeButtonHandle = () => {
+    sendActionButtonEvent('close');
+    onCloseHandle();
   };
 
   return (
@@ -35,7 +64,7 @@ const SkillsBuilderModal = () => {
         size="fullscreen"
         className="skills-builder-modal bg-light-200"
         isOpen
-        onClose={onCloseHandle}
+        onClose={closeButtonHandle}
       >
         <ModalDialog.Hero>
           <ModalDialog.Hero.Background className="bg-primary-500">
@@ -50,13 +79,15 @@ const SkillsBuilderModal = () => {
 
         <ModalDialog.Body>
           <Container size="md" className="p-4.5">
-            <Stepper.Step eventKey={STEP1} title={formatMessage(messages.selectPreferences)}>
-              <SelectPreferences />
-            </Stepper.Step>
+            <Form>
+              <Stepper.Step eventKey={STEP1} title={formatMessage(messages.selectPreferences)}>
+                <SelectPreferences />
+              </Stepper.Step>
 
-            <Stepper.Step eventKey={STEP2} title={formatMessage(messages.reviewResults)}>
-              <ViewResults />
-            </Stepper.Step>
+              <Stepper.Step eventKey={STEP2} title={formatMessage(messages.reviewResults)}>
+                <ViewResults />
+              </Stepper.Step>
+            </Form>
           </Container>
         </ModalDialog.Body>
 
@@ -67,7 +98,7 @@ const SkillsBuilderModal = () => {
             </Button>
             <Stepper.ActionRow.Spacer />
             <Button
-              onClick={() => setCurrentStep(STEP2)}
+              onClick={nextStepHandle}
               disabled={careerInterests.length === 0}
             >
               {formatMessage(messages.nextStepButton)}
@@ -81,7 +112,7 @@ const SkillsBuilderModal = () => {
               {formatMessage(messages.goBackButton)}
             </Button>
             <Stepper.ActionRow.Spacer />
-            <Button onClick={onCloseHandle}>
+            <Button onClick={exitButtonHandle}>
               {formatMessage(messages.exitButton)}
             </Button>
           </Stepper.ActionRow>
--- a/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
+++ b/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
@@ -4,6 +4,7 @@ import {
   IconButton, Icon,
 } from '@edx/paragon';
 import { useIntl } from '@edx/frontend-platform/i18n';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { Close } from '@edx/paragon/icons';
 import { SkillsBuilderContext } from '../../skills-builder-context';
 import { removeCareerInterest } from '../../data/actions';
@@ -12,6 +13,21 @@ import messages from './messages';
 const CareerInterestCard = ({ interest }) => {
   const { formatMessage } = useIntl();
   const { dispatch } = useContext(SkillsBuilderContext);
+
+  const handleRemoveCareerInterest = () => {
+    dispatch(removeCareerInterest(interest));
+
+    sendTrackEvent(
+      'edx.skills_builder.career_interest.removed',
+      {
+        app_name: 'skills_builder',
+        category: 'skills_builder',
+        learner_data: {
+          career_interest: interest,
+        },
+      },
+    );
+  };
 
   return (
     <div className="d-flex justify-content-between align-items-center pb-2 pr-2 pl-4 rounded shadow-sm">
@@ -22,7 +38,7 @@ const CareerInterestCard = ({ interest }) => {
         iconAs={Icon}
         src={Close}
         alt={`${formatMessage(messages.removeCareerInterestButtonAltText)} ${interest}`}
-        onClick={() => dispatch(removeCareerInterest(interest))}
+        onClick={handleRemoveCareerInterest}
       />
     </div>
   );
--- a/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
+++ b/src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
@@ -1,6 +1,7 @@
 import React, { useContext } from 'react';
 import { getConfig } from '@edx/frontend-platform';
 import { useIntl } from '@edx/frontend-platform/i18n';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import {
   Stack, Row, Col, Form,
 } from '@edx/paragon';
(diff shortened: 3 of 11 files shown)


src/skills-builder/skills-builder-modal/select-preferences/CareerInterestSelect.jsx
```diff
@@ -33,6 +42,7 @@ const CareerInterestSelect = () => {
           <JobTitleInstantSearch
             onSelected={handleCareerInterestSelect}
             placeholder={formatMessage(messages.careerInterestInputPlaceholderText)}
+            data-testid="career-interest-select"
```
MaxFrank13, line 45: This is used by the tests to easily change the value of the input

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -50,13 +76,15 @@ const SkillsBuilderModal = () => {
...
+                <SelectPreferences />
+              </Stepper.Step>
 
-            <Stepper.Step eventKey={STEP2} title={formatMessage(messages.reviewResults)}>
-              <ViewResults />
-            </Stepper.Step>
+              <Stepper.Step eventKey={STEP2} title={formatMessage(messages.reviewResults)}>
+                <ViewResults />
+              </Stepper.Step>
+            </Form>
```
MaxFrank13, line 87: Wrapping the Skills Builder in a `Form` component — a minor update that was missed initially, does not effect UI but should have a positive impact on Accessibility

src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
```diff
@@ -12,16 +13,33 @@ import messages from './messages';
...
+    // only setCurrentJobTitle if the user hasn't selected a current job as we don't want to override their selection
+    if (!currentJobTitle) { dispatch(setCurrentJobTitle(value)); }
+
+    sendTrackEvent(
+      `edx.skills_builder.current_job.${value}`,
+      {
+        category: 'skills_builder',
+      },
+    );
+  };
```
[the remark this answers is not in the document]
jsnwesson: Potentially unrelated, but what if the user were to fill a checkbox _before_ searching for their `currentJob`? Would this conflict with any part of the user experience?

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -28,14 +29,39 @@ const SkillsBuilderModal = () => {
     history.goBack();
   };
 
+  const sendActionButtonEvent = (eventSuffix) => {
+    sendTrackEvent(
+      `edx.skills_builder.${eventSuffix}`,
+      {
+        category: 'skills_builder',
+        learnerData: {
+          ...state,
```
jsnwesson, line 38: Given that we're sending all of the `state` when the user clicks `exit` or `close`, do we know what all would be sent other than `currentGoal`, `currentJob`, and `careerInterests`? Do we want to be more explicit?

src/skills-builder/skills-builder-modal/select-preferences/test/SelectPreferences.test.jsx
```diff
@@ -2,8 +2,13 @@ import {
   screen, render, cleanup, fireEvent,
 } from '@testing-library/react';
 import { mergeConfig } from '@edx/frontend-platform';
+import { sendTrackEvent } from '@edx/frontend-platform/analytics';
 import { SkillsBuilderWrapperWithContext, dispatchMock, contextValue } from '../../../test/setupSkillsBuilder';
 
+jest.mock('@edx/frontend-platform/analytics', () => ({
+  sendTrackEvent: jest.fn(),
```
jsnwesson, line 9: Is this relevant if you've got the mock in `setupSkillsBuilder.jsx`?

src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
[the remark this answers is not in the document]
MaxFrank13: Good question! So if they haven't set a `currentJob`, then this will dispatch an action (`setCurrentJobTitle`) to set it to `student` or `looking_for_work`. This will trigger the next question in the form.

If the user were to then set their job from the dropdown menu, that would overwrite `currentJob` to be whatever value they set it to.

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
MaxFrank13: Those are the only values currently being stored in `state`. We could extract them here to be more explicit if we feel it's necessary.

I would think that anything else we end up storing in `state` would also be relevant to the event, but there's no way to be sure. It's easy enough to extract them as we are already extracting `careerInterests` in this same file. I'll make that change 👍

src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
```diff
@@ -35,6 +36,19 @@ const CarouselStack = ({ selectedRecommendations }) => {
     </h3>
   );
 
+  const handleCourseCardClick = (courseKey, productType) => {
+    sendTrackEvent(
+      'edx.skills_builder.recommendation.click',
+      {
+        category: 'skills_builder',
+        page: 'skills_builder',
```
justinhynes, line 44: What is the difference between `category` and `page`? How were these values determined? Should they be the same thing? Should page be something more descriptive like `results`?

src/skills-builder/skills-builder-modal/select-preferences/test/SelectPreferences.test.jsx
MaxFrank13: Good catch! It will work if it's mocked individually in each file. Having it in `setupSkillsBuilder` is actually not doing anything for us now so that can be removed at the very least. I'm not sure how to get the tests to mock using just the setup file. It doesn't seem to pick up on it being a `jest` function

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -1,8 +1,9 @@
 import React, { useState, useContext } from 'react';
 import {
-  Button, Container, Stepper, ModalDialog,
+  Button, Container, Stepper, ModalDialog, Form,
```
hurtstotouchfire, line 3: yaaaaaaaas a real form

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
hurtstotouchfire: yes that was my thinking as well

justinhynes (approved):

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -81,7 +109,7 @@ const SkillsBuilderModal = () => {
               {formatMessage(messages.goBackButton)}
             </Button>
             <Stepper.ActionRow.Spacer />
-            <Button onClick={onCloseHandle}>
+            <Button onClick={exitButtonHandle}>
```
hurtstotouchfire, line 112: What does `esc` do as a keyboard shortcut? I would expect it to fire this action if we were in a modal but I don't think it's a real modal. It's ok if the answer is "it does nothing", I'm just curious.

MaxFrank13: It fires [this function ](https://github.com/openedx/frontend-app-profile/blob/057c351fd0c484ce80481c70397e58609b71d59e/src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx#L64)

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
jsnwesson: The one thing I worry about with being explicit on the properties we want to send out for `exit` and `close` is that _if_ the user decides to leave the quiz _before_ they finish, will there be any complaints from `sendTrackEvent` for having empty/non-existent properties. If you can account for that, this should work.

src/skills-builder/skills-builder-modal/select-preferences/test/SelectPreferences.test.jsx
jsnwesson: Setup is great, when it works the way we want it to. 😆

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
jsnwesson: ooh good point. From an accessibility standpoint, I'd imagine we want it to still close out, given we're treating this like a modal.

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
MaxFrank13: Good catch! Just double checked to be sure that Segment is okay to receive empty strings as values. Each of the values in state default to an empty string, so they will all be defined when we go to `sendTrackEvent` 👍

src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
MaxFrank13: The `page` value comes from the `edx.bi.user.recommended.course.click` event that's being fired in `learner_home`. We wanted to match that one as closely as possible. `category` is used in all of the other events in Skills Builder. That's how they ended up being the same thing. 

In `learner_home`, the `page` is set to `dashboard` so I can see what you mean by setting it to `results` here 🤔

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
MaxFrank13: Yes indeed! The `ModalDialog` component we are using in Paragon has that functionality built-in. It will fire its `onClose` call back function when the `esc` key is pressed or the `X` button clicked

src/skills-builder/skills-builder-modal/select-preferences/CareerInterestCard.jsx
```diff
@@ -13,6 +14,18 @@ const CareerInterestCard = ({ interest }) => {
...
   const { dispatch } = useContext(SkillsBuilderContext);
 
+  const handleRemoveCareerInterest = () => {
+    dispatch(removeCareerInterest(interest));
+
+    sendTrackEvent(
+      'edx.skills_builder.career_interest.removed',
+      {
+        category: 'skills_builder',
+        careerInterest: interest,
```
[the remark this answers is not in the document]
MaxFrank13: We discussed with the team and decided that **yes indeed we do want to match payloads across events**.

I will make these changes 👍

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -66,9 +69,35 @@ const ViewResults = () => {
...
+    }
+    setSelectedRecommendations(newlySelectedRecommendations);
+  }, [productRecommendations, selectedJobTitle, selectedRecommendations, sendRecommendationShownEvent]);
+
+  const handleJobTitleChange = (e) => {
+    const { value } = e.target;
+    setSelectedJobTitle(value);
+    // The is_default value will be set to false for any selections made by the user
+    sendRecommendationShownEvent(value, false);
+  };
```
MaxFrank13, line 100: Newly added code for the `edx.skills_builder.recommendation.shown` event that will fire in two places:

1. Once for the default selection that is made by the app when the recommendations are first loaded. We default this selection to whichever set of recommendations is returned first in the list. For this event, the `is_default` field is set to `true`.
2. And then whenever a user makes a selection through the `onChange` handle for the `SelectableBox`(radio button) components. The `is_default` field is set to `false` whenever the user makes this selection.

The utility function that is re-used to fire both events is wrapped in a `useCallback` hook to avoid issues with the `useEffect` dependency array.

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -106,7 +135,8 @@ const ViewResults = () => {
         <RelatedSkillsSelectableBoxSet
           jobSkillsList={jobSkillsList}
           selectedJobTitle={selectedJobTitle}
-          onChange={(e) => setSelectedJobTitle(e.target.value)}
+          // TODO: add handle function to call setSelectedJobTitle and the 'shown' event
```
justinhynes, line 138: Do we still need this TODO? It looks like we may have implemented it?

src/skills-builder/skills-builder-modal/SkillsBuilderModal.jsx
```diff
@@ -28,14 +29,42 @@ const SkillsBuilderModal = () => {
     history.goBack();
   };
 
+  const sendActionButtonEvent = (eventSuffix) => {
+    sendTrackEvent(
+      `edx.skills_builder.${eventSuffix}`,
+      {
+        app_name: 'skills_builder',
```
justinhynes, line 36: Sorry to re-hash this. Should the "app_name" here be "profile"? Since it's coming from the Profile MFE? This might be able to help us differentiate metrics from this features life as part of the profile MFE (compared to the dedicated MFE we expect to build for it). Am I just thinking about this wrong? I know we talked about this during standup the other day but I think I'm still a little fuzzy here

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
MaxFrank13: Good catch!

PR #744, opened 2023-05-10
fix: adding configure component to filter jobs

MaxFrank13:
This PR adds filtering for the job titles we receive from Algolia. This is done using the `Configure` component supplied by the `react-instantsearch-hooks-web` package.

[Documentation on `Configure`](https://www.algolia.com/doc/api-reference/widgets/configure/react-hooks/)

The filter had to be added to Algolia as well:

Screenshot 2023-05-10 at 4 16 04 PM

### JIRA
[APER-2405](https://2u-internal.atlassian.net/browse/APER-2405)

### DIFF
--- a/src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
+++ b/src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
@@ -5,7 +5,7 @@ import {
 } from '@edx/paragon';
 import { useIntl } from '@edx/frontend-platform/i18n';
 import { sendTrackEvent } from '@edx/frontend-platform/analytics';
-import { InstantSearch } from 'react-instantsearch-hooks-web';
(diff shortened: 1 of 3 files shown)


src/skills-builder/skills-builder-modal/select-preferences/JobTitleSelect.jsx
```diff
@@ -52,6 +52,7 @@ const JobTitleSelect = () => {
           {formatMessage(messages.jobTitlePrompt)}
         </h4>
         <InstantSearch searchClient={searchClient} indexName={getConfig().ALGOLIA_JOBS_INDEX_NAME}>
+          <Configure filters="b2c_opt_in:true" />
```
justinhynes, line 55: Is this the first selection box? Like, where we ask what job the person _currently_ has? If so, I don't think we should be filtering here. We should only filter on the when we ask "what are you interested in pursuing". Otherwise, we might not be able to capture the user's information, because their current job might not be in our list.

PR #750, opened 2023-05-16
fix: reduce event payload

MaxFrank13:
### Context

We noticed that two of our events were not firing when we expected them to:
- `edx.skills_builder.recommendation.shown`
- `edx.skills_builder.recommendation.click`

After checking the [Segment documentation](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/troubleshooting/#is-there-a-size-limit-on-requests), we learned that there is indeed a limit put on the payload for the event.

Check the [wiki page for more details](https://2u-internal.atlassian.net/wiki/spaces/ENGAGE/pages/445612347/Segment+dropping+events+with+large+payload).

### Description

This PR modifies the payload of these two events so that they don't include nearly as much data. Previously, we were nesting the entire Algolia response into the payload. We did this so that we have a reference to each recommendation that is shown to the user. The event is called once the default recommendations are loaded. It is also called as the user navigates their selections. There is an `is_default` property for this event to help provide a denominator when it comes to aggregate this data.

Instead of the entire Algolia response, we are instead extracting the `active_run_key` as `course_key` and the `title` of the course. See below:

#### Previous payload
```
...unchanged properties,
selected_recommendations: {
   job_id: 12345
   job_name: Software Engineer
   recommendations: {
       course: [
           {
            <Insert response from Algolia — about 47 attributes, including title and active_run_key>
           },
           {
            <Insert response from Algolia — about 47 attributes, including title and active_run_key>
           }
       ] 
   }
}
```
#### New payload
```
...unchanged properties,
selected_recommendations: {
   job_id: 12345
   job_name: Software Engineer
   course_keys: [
      {
       title: 'Learning new things',
       course_key: '<insert-ID-here>`
      },
      {
       title: 'Learning more things',
       course_key: '<insert-ID-here>`
      }
   ] 
}
```

### DIFF
--- a/src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
+++ b/src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
@@ -7,8 +7,12 @@ import messages from './messages';
 
 const CarouselStack = ({ selectedRecommendations }) => {
   const { formatMessage } = useIntl();
-  const { name: jobName, recommendations } = selectedRecommendations;
+  const { id: jobId, name: jobName, recommendations } = selectedRecommendations;
   const productTypeNames = Object.keys(recommendations);
+  const courseKeys = recommendations.course?.map(rec => ({
+    title: rec.title,
+    courserun_key: rec.active_run_key,
+  }));
 
   const normalizeProductTypeName = (productType) => {
     // If the productType is more than one word (i.e. boot_camp)
@@ -43,9 +47,13 @@ const CarouselStack = ({ selectedRecommendations }) => {
         app_name: 'skills_builder',
         category: 'skills_builder',
         page: 'skills_builder',
-        course_key: courseKey,
+        courserun_key: courseKey,
         product_type: productType,
-        selected_recommendations: selectedRecommendations,
+        selected_recommendations: {
+          job_id: jobId,
+          job_name: jobName,
+          courserun_keys: courseKeys,
+        },
       },
     );
   };
--- a/src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
+++ b/src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
@@ -1,5 +1,5 @@
 import React, {
-  useCallback, useContext, useEffect, useState,
+  useContext, useEffect, useState,
 } from 'react';
 import {
   Stack, Row, Alert, Spinner,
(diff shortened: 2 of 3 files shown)


src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
```diff
@@ -45,7 +49,11 @@ const CarouselStack = ({ selectedRecommendations }) => {
         page: 'skills_builder',
         course_key: courseKey,
         product_type: productType,
-        selected_recommendations: selectedRecommendations,
+        selected_recommendations: {
+          job_id: jobId,
+          job_name: jobName,
+          course_keys: courseKeys,
+        },
```
MaxFrank13, line 56: I wanted the fields here to be consistent with the rest of the event data so they have been `snake_cased`

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -58,9 +58,24 @@ const ViewResults = () => {
...
+          job_id: results[0].id,
+          job_name: results[0].name,
+          /* We extract the title and course key into an array of objects */
+          course_keys: results[0].recommendations.course?.map(rec => ({
+            title: rec.title,
+            course_key: rec.active_run_key,
+          })),
+        },
+        is_default: true,
+      });
```
MaxFrank13, line 78: While this may be making the code a little less DRY, I found this implementation to be a lot less clunky. Especially in terms of getting it to fire _only once_ when the recommendations first load. Also, the previous implementation was causing React to loop it's `useEffect` when it wasn't necessary, making the application a little less performant.

Moreover, all of the event code is eventually going to live somewhere else and I figure that will be the best time to refactor things further. That work is captured by [this ticket](https://2u-internal.atlassian.net/browse/APER-2392) .

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -58,9 +58,24 @@ const ViewResults = () => {
...
+        selected_recommendations: {
+          job_id: results[0].id,
+          job_name: results[0].name,
+          /* We extract the title and course key into an array of objects */
+          courserun_keys: results[0].recommendations.course?.map(rec => ({
+            title: rec.title,
+            courserun_key: rec.active_run_key,
+          })),
+        },
+        is_default: true,
```
justinhynes, line 77: Should `is_default` always be hardcoded to `true` here? In the code below it wasn't hardcoded.

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
justinhynes: Can we put this comment in the code somewhere? This is interesting context that explains why some of the code is duplicated. I could see someone coming through and trying to refactor this and not knowing the historical context.

justinhynes (approved):

PR #751, opened 2023-05-17
fix: close and exit buttons

MaxFrank13:
### Description
This modifies the functionality of the "Close/X" and "Exit" buttons in the Skills Builder. They previously were utilizing the client's history to go back when clicked. 

There was an oversight when first building this out — the Skills Builder is always opened up in a separate tab, so there is no history to go back to. Instead, these buttons have been implemented as `Hyperlink` components that direct the user to a set destination: `https://edx.org/search` for production. We use the `SEARCH_CATALOG_URL` environment variable to do this. 

This variable [comes from the `common/prod_config.yml` file](https://github.com/edx/edx-internal/blob/master/frontends/common/prod_config.yml). There is one for Stage as well. Because it's in the common config, we do not need to update `edx-internal` to have access to it. The link for dev has been set to `http://localhost:18000/courses` based on [prior art in `frontend-app-learning`](https://github.com/openedx/frontend-app-learning/blob/master/.env.development#L35).

There has also been a modification to the UI to remove the "Go Back" button on the first step of the Skills Builder. It was not in the initial mock-up and doesn't serve much of a purpose.

### Screenshot
Screenshot 2023-05-17 at 12 51 22 PM


src/index.jsx
```diff
@@ -58,6 +58,7 @@ initialize({
         ALGOLIA_JOBS_INDEX_NAME: process.env.ALGOLIA_JOBS_INDEX_NAME || null,
         ALGOLIA_PRODUCT_INDEX_NAME: process.env.ALGOLIA_PRODUCT_INDEX_NAME || null,
         ALGOLIA_SEARCH_API_KEY: process.env.ALGOLIA_SEARCH_API_KEY || null,
+        SEARCH_CATALOG_URL: process.env.SEARCH_CATALOG_URL || null,
```
justinhynes, line 61: nit: The name `Catalog` has a loaded meaning, as the alternative name for the Course-Discovery IDA is "Catalog". I think it's a bit clearer if the name of this setting refers to the marketing site in some way? 

This is not a hill I'm going to die on and feel free to ignore me if you feel otherwise.

justinhynes (approved):

src/index.jsx
MaxFrank13: This makes sense. We can't modify the name as it appears in the `.env` files as that is determined by the common config. But we can modify it here where you have highlighted so that whenever it's referenced in the app, the name is `MARKETING_SITE_SEARCH_URL`. I've made those changes 👍

PR #756, opened 2023-05-24
fix: adjust header height and content when window is medium in width or less

jsnwesson:
# Context
A UI bug was reported ([APER-2341](https://2u-internal.atlassian.net/browse/APER-2341)) where when the view is a small enough width, the header's height will take up a large portion and thus makes the Course Recommendations in Skills Builder not as apparent. 

# Solution
When the view is of medium size (992px) or less, decrease the height of the header and simplify the heading. 

This will also change heading content to "edX Skills builder" when screen is medium or smaller.

One other addition is that when the viewport is medium or less, the stepper's header will switch to a compact view (only display current step and "Step # out of #") 

Screenshot 2023-06-02 at 3 15 03 PM


src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
```diff
@@ -3,3 +3,7 @@
     color: #ffffff;
   }
 }
+
+.med-min-height {
+  min-height: 5rem;
```
MaxFrank13, line 8: We can use SCSS variables for these values as a best practice. I don't think it really has a huge effect here but these values could be changed to:
- `#ffffff` to `$white`
- `5rem` to `map-get($spacers, 6)`

I think we just missed the first one in a previous PR.

jsnwesson: Awesome, I'll add these suggestions!

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
```diff
@@ -1,11 +1,22 @@
 import React from 'react';
 import { useIntl } from '@edx/frontend-platform/i18n';
+import { breakpoints, useMediaQuery } from '@edx/paragon';
 import edXLogo from '../images/edX-logo.svg';
 import messages from './messages';
 
 const SkillsBuilderHeader = () => {
   const { formatMessage } = useIntl();
+  const isMedium = useMediaQuery({ maxWidth: breakpoints.medium.maxWidth });
```
MaxFrank13, line 9: Seeing as we are doing this in the parent component as well (`SkillsBuilderModal`), what do you think about passing `isMedium` down as a prop to `SkillsBuilderHeader`? The value already exists in `SkillsBuilderModal` so this would eliminate the need to import the hook in two places.

src/skills-builder/skills-builder-header/messages.js
```diff
@@ -11,6 +11,11 @@ const messages = defineMessages({
     defaultMessage: 'Let edX be your guide',
     description: 'Subheading to the Skills Builder title in the header component',
   },
+  skillsBuilderHeaderTitleIsMedium: {
+    id: 'skills.builder.header.title.is.Medium',
```
MaxFrank13, line 15: nit: should `Medium` be lowercase?

MaxFrank13 (commented):
One nit and one suggestion. Let me know what you think! Thanks!

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
```diff
@@ -1,5 +1,9 @@
 .skills-builder-modal {
   button[aria-label="Close"][type="button"]{
-    color: #ffffff;
+    color: $white;
   }
 }
+
+.med-min-height {
+  min-height: map-get($spacers, 6);
+}
```
justinhynes, line 9: Tangential question: Is this howe we handled similar problems in other MFEs? Is this considered "prior art" for restraining views for smaller screens? Can a similar effect be supported using built-in bootstrap classes? I've always heard in the past that adding Sass/CSS classes should be a last ditch effort. I thought we could also handle breakpoints/sizing through bootstrap (even though I'm not well versed in it).

I know that adding one new CSS class/definition probably isn't a big deal but, as I understand it, we try as best as possible to keep bundle sizes as small as possible, which means as little custom CSS as required.

justinhynes (approved):

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
jsnwesson: Good catch!

src/skills-builder/skills-builder-header/messages.js
jsnwesson: Good catch on this, too! Didn't intend for that to be uppercase

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
justinhynes: Gotcha! Yeah, this was more for my personal knowledge/learning than "I think this is a problem" :)

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
```diff
@@ -1,25 +1,39 @@
...
-const SkillsBuilderHeader = () => {
+const SkillsBuilderHeader = ({ isMedium }) => {
   const { formatMessage } = useIntl();
 
+  if (isMedium) {
+    return (
+      <div className="d-flex">
+        <h3 className="mb-0 text-white">
+          {formatMessage(messages.skillsBuilderHeaderTitleIsMedium)}
+        </h3>
```
justinhynes, line 15: Now that I'm taking a second look at this, I'm worried if we have a potential a11y issue here.

In the previous version, the title was displayed inside an `H1`. In this updated version, we have an `H3` here. Should this be an `H1` _styled like_ an `H3` (adding the `h3` declaration to the `className`)?

If the title of the header should be an H1, shouldn't the mobile-sized version also be an H1? 

More: I was taught that we should only have a single H1 on the DOM. This looks like we may potentially have NO H1 if we default to an H3?

src/skills-builder/skills-builder-header/SkillsBuilderHeader.jsx
```diff
@@ -1,25 +1,39 @@
...
       <div className="w-100 ml-5">
-        <h1 className="h1 text-warning-300">
+        <h1 className="text-warning-300">
           {formatMessage(messages.skillsBuilderHeaderTitle)}
         </h1>
-        <p className="h2 text-white">
+        <h2 className="text-white">
           {formatMessage(messages.skillsBuilderHeaderSubheading)}
-        </p>
+        </h2>
```
justinhynes, line 29: Similarly, related to my above comment, should this really be an h2? Previously, it was just text _styled_ like an H2, which (I thought) I had been told was preferable than defining header levels directly.

PR #757, opened 2023-05-25
fix: fallback src typo for course cards

MaxFrank13:
There was a typo in the attributes for the `RecommendationCard` component causing there to be no fallback image. This was making the course cards look awkward while waiting for the images to load.


justinhynes (approved):
Good catch. LGTM, checked this against examples in Paragon and this should fix it up.

PR #763, opened 2023-05-31
[Do Not Merge] build: demonstrate using localhost.stage.edx.org

adamstankiewicz:
# PoC Demo of MFE Development w/o Devstack 🤘 

## Purpose
To increase the FE development iteration speed. To reduce the cognitive load required to make changes so developers can focus on delivering customer value. 

MFE development relies by default on devstack-based services. To run _any_ individual devstack service a developer has to run almost _every_ devstack service - this requires a lot of effort and cognitive load to configure and maintain. Frontend engineers would benefit from an option to run no/fewer devstack services.

A positive side-effect is the data in stage will become better maintained through the normal course of developer-effort.

## HowTo
- make a local domain/dns change
  - `sudo nano /etc/hosts`
  - add `127.0.0.1 localhost.stage.edx.org`
- make config tweaks in line with this PR
- https://localhost.stage.edx.org:1995

## Stage Resources

- Common https://github.com/edx/edx-internal/blob/master/frontends/common/stage_config.yml
- App Example https://github.com/edx/edx-internal/blob/master/frontends/frontend-app-admin-portal/stage_config.yml

## Additional Supporting Context
We've made some additional config tweaks on stage to support this. Some are temporary until we figure out a cleaner path.

- CORS origin allowlist for `localhost.stage.edx.org`
  - for testing, it's `CORS_ORIGIN_ALLOW_ALL: true` for now on courses.stage.edx.org until a better strategy is figured out.
- `CSRF_TRUSTED_ORIGINS` needs `.edx.org:{PORT}`

## Future Improvements & Questioins
- allow python IDAs to run w/ stage dependencies (ie run _less_ devstack)
  - devstack decoding of stage JWT tokens?


johnnagro:
Starting to formalize docs here https://2u-internal.atlassian.net/wiki/spaces/FEDX/pages/514162781/MFE+development+without+devstack

justinhynes:
I followed up with @adamstankiewicz and this PR can be closed for now.

PR #771, opened 2023-06-09
fix: switch out class overwrite in SCSS for an aria-label selector

jsnwesson:
# Context: 

In [APER-2341](https://github.com/openedx/frontend-app-profile/pull/756) I originally overwrote a class's CSS styling, which is not considered best practice as the class (`pgn__modal-close-container`) could be renamed at any point and the overwrite would create a breaking change. 

This PR replaces that implementation by referring to the `aria-label` of the button and it still achieves the same effect. 

# Screenshots

### Prior to any styling changes, the close button's X was slightly higher and not aligned to the heading, which for medium viewports (or less) we want to keep them aligned. 
Screenshot 2023-06-09 at 11 20 09 AM

### After applying the styling, the X is aligned with the heading. 
Screenshot 2023-06-09 at 11 19 48 AM

### DIFF
--- a/src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
+++ b/src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
@@ -1,16 +1,14 @@
-.skills-builder-modal {
-  button[aria-label="Close"][type="button"]{
-    color: $white;
-  }
+button[aria-label="Close"][type="button"]{
+  color: $white;
 }
-
 
 $breakpoint-medium: 992px;
 @media (max-width: $breakpoint-medium) {
   .med-min-height {
     min-height: map-get($spacers, 6);
   }
-  .pgn__modal-close-container {
-    top: 1rem;
+  button[aria-label="Close"][type="button"]{
+    position: relative;
+    top: 0.5rem;
   }
 }


src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
```diff
@@ -1,16 +1,14 @@
-.skills-builder-modal {
-  button[aria-label="Close"][type="button"]{
-    color: $white;
-  }
+button[aria-label="Close"][type="button"]{
+  color: $white;
```
MaxFrank13, line 2: Nit: Is there a reason why this selector has been un-nested from the `skills-builder-modal` class selector?

MaxFrank13 (approved):
Had a question but I don't believe anything needs to be changed. LGTM!

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
jsnwesson: Good point. I had a reason at first, and now I realize that the point of wrapping it with the skills builder class selector was so it only affected this component and no other potential modal usages. I'll revert it!

PR #775, opened 2023-06-12
Query string to act as switch for lines of business

MaxFrank13:
This PR adds the ability to adjust which lines of business the Skills Builder will return recommendations for. This can be set using the query string in the URL.

The hook we are using expects the query string to be formatted as one key-value pair. The key is `product_types` and the value is a comma separated list of product types.

`product_types=<line_of_business>,<line_of_business>`

For example:
`skills.edx.org?product_types=boot_camp,course,2u_degree` 

This will tell the app to show recommendations for Boot Camps, Open Courses, and 2U Degrees.

If a query string is not provided or is formatted incorrectly, the app will default to only showing Open Courses.

https://github.com/openedx/frontend-app-profile/assets/92897870/bc220c37-c1c6-42dc-a955-bc2ba7a55276


src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
```diff
@@ -14,3 +14,7 @@ $breakpoint-medium: 992px;
     top: 1rem;
   }
 }
+
+.chip-max-width {
+  max-width: 16rem;
```
MaxFrank13, line 19: This is to fix an issue where the chips were forcing a wider width on the `SelectableBox` components. This will truncate the text if it is too long.

src/skills-builder/skills-builder-modal/view-results/CarouselStack.jsx
```diff
@@ -21,7 +21,7 @@ const CarouselStack = ({ selectedRecommendations }) => {
       const splitStrings = productType.split('_');
 
       // map through the array and normalize each string (i.e. ['Boot', 'Camp'])
-      const normalizeStrings = splitStrings.map(word => word[0].toUpperCase() + word.slice(1).toLowerCase());
+      const normalizeStrings = splitStrings.map(word => word[0].toUpperCase() + word.slice(1));
```
MaxFrank13, line 24: Previously, "2U_degree" would appear as "2u Degree" in the UI. This makes it so it will read 2U Degree in the UI without effecting the other product types (lines of business)

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -77,11 +55,13 @@ const ViewResults = () => {
...
     };
-    getRecommendations()
+
+    getAllRecommendations()
       .catch(() => {
         setFetchError(true);
         setIsLoading(false);
       });
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [careerInterests, jobSearchIndex, productSearchIndex]);
```
MaxFrank13, line 65: This has been refactored into the `service.js` file which is more in line with out best practice. It also cleans up this file a little bit.

src/skills-builder/skills-builder-modal/view-results/ViewResults.jsx
```diff
@@ -27,35 +28,12 @@ const ViewResults = () => {
...
 
-          // replace all white spaces with an underscore
-          const formattedProductType = productType.replace(' ', '_');
-
-          // add a new key to the recommendations object and set the value to the response
-          data.recommendations[formattedProductType] = response;
-        }));
-
-        return data;
-      }));
```
MaxFrank13, line 58: This has been refactored into the `service.js` file. This is more in line with our best practices and also cleans up the file a little bit.

src/skills-builder/utils/search.js
MaxFrank13, line 1: Nit: this was previously a `.jsx` file when it technically didn't need to be. This doesn't have a huge effect on things but it can be nice to know which files have React in them and which do not.

src/skills-builder/skills-builder-modal/skillsBuilderModal.scss
MaxFrank13: It was not logged for the `SelectableBox` component. That was something I picked up on in a meeting with Gabe.

However, as it turns out, this also fixes the issue we were seeing where Cards in the carousel were getting melded together. That [bug was logged here](https://2u-internal.atlassian.net/browse/APER-2466).

justinhynes (approved):

justinhynes:
LGTM, please remember to squash your commits before merging :)

PR #788, opened 2023-06-27
refactor: remove Skills Builder from Palm

MaxFrank13:
This PR will remove the B2C Skills Builder, any associated tests, and any dependencies that will no longer be used, from the Profile MFE. This is part of work Aperture is doing to port Skills Builder over to its own MFE. The Skills Builder is not currently being used anywhere so it is safe to remove.

This feature is currently disabled as we prepare for the next wave of experimentation. This change to Palm will differ from what appears in the master branch, however, soon the Skills Builder will be removed from master as well.


.env
```diff
@@ -25,9 +25,4 @@ FAVICON_URL=''
 COLLECT_YEAR_OF_BIRTH=true
 APP_ID=''
 MFE_CONFIG_API_URL=''
-ENABLE_SKILLS_BUILDER=''
 ENABLE_SKILLS_BUILDER_PROFILE=''
```
jsnwesson, line 28: Can you give context on why we would keep this flag?

jsnwesson:
left one comment about keeping the feature flag `ENABLE_SKILLS_BUILDER_PROFILE`, but otherwise this looks good!

.env
MaxFrank13: That is the flag used by the Profile page to decide whether or not to render the learner's goal. I think we still want this flag unless there are plans to remove the feature altogether

justinhynes (approved):

PR #790, opened 2023-06-28
feat: Pact Consumer Contract Testing on LMS Account API

swayamrana:
# Pact Consumer Contract Testing on LMS Account API

## Context

This PR creates consumer contracts between the Profile MFE and the LMS that will help define expectations between each service. These expectations will be coordinated through the Pact stub server in a Docker container. The next steps include implementing provider tests and verifying requests & responses in the User API repository listed [here](https://github.com/openedx/edx-platform/tree/96f13978720d3fc8760aa71e435dd7dad2782dcf/openedx/core/djangoapps/user_api). Ultimately, this stub server will replicate the entire LMS architecture in local development, increasing efficiency and improving developer experience as a whole.

## Decision

I chose two basic Account API mock responses to stub out for the contract: A HTTP 200 response and a HTTP 404 Response. The contract's expected output is a `.json` file that includes the name, bio, email, etc. of the user, if the username is found. 

## Consequences

In the future, more cases will have to be documented that cover the varying LMS responses, which test different HTTP responses and distinct account information. However, the current test file provides a good starting point that leaves room for expansion and improvement. 

## References

[This](https://github.com/openedx/frontend-app-profile/compare/master...djoy/dev_auth_service_pact) PR gives the average developer who wants to read up on Pact documentation an alternative source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
-    "test": "TZ=UTC fedx-scripts jest --coverage --passWithNoTests"
+    "test": "TZ=UTC fedx-scripts jest --coverage --passWithNoTests",
+    "stubs": "pact-stub-service ./src/pacts/frontend-app-profile-edx-platform.json --port 18000"
   },
   "bugs": {
     "url": "https://github.com/openedx/frontend-app-profile/issues"
@@ -37,6 +38,7 @@
     "@fortawesome/free-regular-svg-icons": "5.15.4",
     "@fortawesome/free-solid-svg-icons": "5.15.4",
     "@fortawesome/react-fontawesome": "0.2.0",
+    "@pact-foundation/pact": "^11.0.2",
     "classnames": "2.3.2",
     "core-js": "3.31.0",
     "history": "5.3.0",
--- a/src/pacts/frontend-app-profile-edx-platform.json
+++ b/src/pacts/frontend-app-profile-edx-platform.json
@@ -1,0 +1,77 @@
+{
+  "consumer": {
+    "name": "frontend-app-profile"
+  },
+  "interactions": [
+    {
+      "description": "A request for user's basic information",
+      "providerStates": [
+        {
+          "name": "Account and user's information does not exist"
+        }
+      ],
+      "request": {
+        "method": "GET",
+        "path": "/api/user/v1/accounts/staff_not_found"
+      },
+      "response": {
+        "status": 404
+      }
+    },
+    {
+      "description": "A request for user's basic information",
+      "providerStates": [
+        {
+          "name": "I have a user's basic information"
+        }
+      ],
+      "request": {
+        "method": "GET",
+        "path": "/api/user/v1/accounts/staff"
+      },
+      "response": {
+        "body": {
+          "bio": "This is my bio",
+          "country": "ME",
+          "dateJoined": "2017-06-07T00:44:23Z",
+          "email": "staff@example.com",
+          "isActive": true,
+          "name": "Lemon Seltzer",
+          "username": "staff",
+          "yearOfBirth": 1901
+        },
+        "headers": {
+          "Content-Type": "application/json"
+        },
+        "matchingRules": {
+          "body": {
+            "$": {
+              "combine": "AND",
+              "matchers": [
+                {
+                  "match": "type"
+                }
+              ]
+            }
+          }
+        },
+        "status": 200
+      }
+    }
+  ],
+  "metadata": {
+    "pact-js": {
+      "version": "11.0.2"
+    },
+    "pactRust": {
+      "ffi": "0.4.0",
+      "models": "1.0.4"
+    },
+    "pactSpecification": {
+      "version": "3.0.0"
+    }
+  },
+  "provider": {
+    "name": "edx-platform"
+  }
+}
--- a/src/profile/data/pact-profile.test.js
+++ b/src/profile/data/pact-profile.test.js
@@ -1,0 +1,80 @@
+// This test file simply creates a contract that defines
+// expectations and correct responses from the Pact stub server.
+
+import path from 'path';
+
+import { PactV3, MatchersV3 } from '@pact-foundation/pact';
+
+import { initializeMockApp, getConfig, setConfig } from '@edx/frontend-platform';
+import { getAccount } from './services';
+
+const expectedUserInfo200 = {
+  username: 'staff',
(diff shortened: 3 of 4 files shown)


src/pacts/frontend-app-profile-edx-platform.json
```diff
@@ -0,0 +1,117 @@
+{
```
timmc-edx, line 1: I'm guessing this file is an unintended duplicate?

swayamrana: Yeah, seems like it generated two pacts: One located in the src folder and one in the entire directory.

timmc-edx: The files are a bit different, with the `staff_not_found` interaction having moved, and a space character having snuck into one of the URLs.

timmc-edx: (`src/pacts/frontend-app-profile-edx-platform.json` was the one with the space char, so that's all good now)

swayamrana: Should i move back the ```staff_not_found``` interaction back? The file was auto-generated from the pact package so a bit confused on that.

package.json
```diff
@@ -14,7 +14,8 @@
     "lint": "fedx-scripts eslint --ext .js --ext .jsx .",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "fedx-scripts webpack-dev-server --progress",
-    "test": "TZ=UTC fedx-scripts jest --coverage --passWithNoTests"
+    "test": "TZ=UTC fedx-scripts jest --coverage --passWithNoTests",
+    "stubs": "$(find . -name pact-stub-service | grep -e 'bin/pact-stub-service$' | head -n 1) ./pacts/frontend-app-profile-edx-platform.json --port 18000"
```
timmc-edx, line 18: When I try to run this, it's looking for a Ruby installation (which I don't have). Also, I know it's from the official docs, but that find/grep/head pipeline seems like it would randomly find the wrong file:

```
timmc@edx-timmc:~/edx-repos/frontend-app-profile$ find . -name pact-stub-service | grep -e 'bin/pact-stub-service$'
./node_modules/@pact-foundation/pact-core/standalone/linux-x64-1.92.0/pact/lib/vendor/ruby/2.4.0/gems/pact-mock_service-3.9.0/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/linux-x64-1.92.0/pact/lib/vendor/ruby/2.4.0/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/linux-x64-1.92.0/pact/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/darwin-1.92.0/pact/lib/vendor/ruby/2.4.0/gems/pact-mock_service-3.9.0/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/darwin-1.92.0/pact/lib/vendor/ruby/2.4.0/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/darwin-1.92.0/pact/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/win32-1.92.0/pact/lib/vendor/ruby/2.4.0/gems/pact-mock_service-3.9.0/bin/pact-stub-service
./node_modules/@pact-foundation/pact-core/standalone/win32-1.92.0/pact/lib/vendor/ruby/2.4.0/bin/pact-stub-service
./node_modules/.bin/pact-stub-service
```

I don't really understand Node's path management, but when I experimented with the following replacement it seemed to work:

```suggestion
    "stubs": "pact-stub-service ./pacts/frontend-app-profile-edx-platform.json --port 18000"
```

...even though I don't have Ruby installed. I think it's using a Dockerized version when called this way.

package.json
```diff
@@ -37,7 +38,8 @@
     "@fortawesome/free-regular-svg-icons": "5.15.4",
     "@fortawesome/free-solid-svg-icons": "5.15.4",
     "@fortawesome/react-fontawesome": "0.2.0",
-    "algoliasearch": "4.17.2",
+    "@pact-foundation/pact": "^11.0.2",
+    "algoliasearch": "4.18.0",
```
timmc-edx, line 42: Is this package upgrade required?

swayamrana: I don't think so, it might have automatically updated my algoliasearch package when tinkering with the dependencies. I'll set it to 4.17.2 and see how it runs.

davidjoy: In terms of keeping commits/PRs focused on a particular set of changes, I'd revert this line (which I think is what Tim's implying) and the associated changes in package-lock.json.  

The easiest way to do that is just edit it back to 4.17.2 in this file manually, then run `npm i` to get the package-lock.json correct.  Editing package-lock.json manually is never a good idea!

davidjoy (approved):
Great first set of tests!  Can't wait for the next part. 😄 

I have a few small suggestions below, but pending that this is ready to merge.  I'm going to pre-emptively approve; I'd minimally address the algolia version number thing, and if you agree with the stylistic suggestion, that too.

src/profile/data/pact-profile.test.js
```diff
@@ -0,0 +1,77 @@
...
+  consumer: 'frontend-app-profile',
+  provider: 'edx-platform',
+});
+
+describe('getAccount for one username', () => {
+  beforeAll(async () => {
+    initializeMockApp();
+  });
+  it('returns a HTTP 200 and user information', async () => {
+    const usernameEdpt200 = 'staff';
```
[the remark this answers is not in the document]
swayamrana: I renamed Endpoint as Edpt in my variable names, I'll go back and change that to the original. In my code, sometimes I tend to shrink down certain words to save space, and end up making everything pretty ambiguous. Will try my best to avoid that in the future! 👍

src/profile/data/pact-profile.test.js
```diff
@@ -0,0 +1,77 @@
+import path from 'path';
```
timmc-edx, line 1: I think it would be good to add a file-level comment explaining that this isn't actually a test, that it just generates the contract files.

timmc-edx (approved):
One suggestion, but looks good! From discussion, probably worth adding an ADR to explain how this work is intended to be expanded on.

(One nit: Commit type should probably be `feat`, rather than `refactor`.)

davidjoy:
> One suggestion, but looks good! From discussion, probably worth adding an ADR to explain how this work is intended to be expanded on.
> 
> (One nit: Commit type should probably be `feat`, rather than `refactor`.)

I might have just gone with `test` since it's not actually a new feature in the app. 😄

timmc-edx:
It's using the testing mechanism, but it doesn't actually perform testing, and the goal is to make a stub server that makes development easier. (Not a user-facing feature, though.)

PR #868, opened 2023-10-12
build: Profile Plugin Page

jsnwesson:
This PR directly inserts the Frontend Plugin Framework that originally lived in Frontend Platform. 

This also includes:
- a Profile page that will be used as a Plugin


src/index.jsx
```diff
@@ -27,6 +28,11 @@ import Head from './head/Head';
 
 import AppRoutes from './routes/AppRoutes';
 
+const RealFooter = () => {
```
MaxFrank13, line 31: NIT: Maybe consider calling this `renderFooter` or something like that? In my opinion, it feels like "real" implies that there is a fake or mocked one as well

MaxFrank13 (approved):
LGTM! One nit pick but not blocking

src/index.jsx
jsnwesson: Yup, makes sense!

PR #870, opened 2023-10-13
fix: fit & finish profile card plugin 

jsnwesson:
This PR will design the Profile Plugin into a card shape that mimics as closely as possible to the mock design attached in [APER-2965](https://2u-internal.atlassian.net/browse/APER-2965).


src/profile/forms/PluginCountry.jsx
cdeery, line 1: nice job refactoring this out of the editable component

cdeery (approved):
looks good

PR #897, opened 2023-10-24
fix: hyperlink in plugin page should redirect to user profile

jsnwesson:
This PR will add the user's full name to the Profile Plugin page, as well as fix the hyperlink in the plugin to send the user to their Profile page.


cdeery (approved):
Looks good. Nice job clearing out the hard-coding!

PR #913, opened 2023-11-14
fix: "view public profile" link sends parent frame to URL

jsnwesson:
Context:
The Profile plugin component has a `view public profile` hyperlink, then when clicked would redirect the page _inside the plugin_ rather than the parent frame (AKA Host MFE).

This PR fixes that bug by adding `target=_parent` to the Hyperlink tag.


cdeery (approved):
Looks great!

PR #920, opened 2023-11-22
fix: remove scrollbars

MaxFrank13:
This adds `overflow: hidden;` to the wrapper for the Plugin to hide scrollbars that will flicker when resizing.


cdeery (approved):
Looks good! I hope that clears up the problem!

PR #1000, opened 2024-04-04
feat: replace frontend-build with vite

bradenmacdonald:
This is a proof of concept that shows how easily [`@openedx/frontend-build`](https://github.com/openedx/frontend-build/), a custom wrapper around webpack and babel, can be replaced with [Vite](https://vitejs.dev/).

As far as I can tell, Vite supports all the functionality we need from `frontend-build`, _plus_ has much better TypeScript integration, is much faster, and most of all requires less maintenance.

The [various `webpack.*.config` file in frontend-build](https://github.com/openedx/frontend-build/tree/master/config) have something like 39+187+213 = 439 lines of code (+41 loc for babel.config.js), whereas the equivalent `vite.config.ts` is only ~48 lines of code - that's 10x less configuration to maintain.

Notes:
* This works "out of the box" with `tutor-mfe` and `frontend-platform`; in fact this PR is the only thing necessary to start using Vite for builds. (Though of course things could be improved even further by modifying tutor and `frontend-platform` accordingly)
* I'm not sure about i18n message builds nor New Relic integration - those may require additional work

## Metrics

Table of metrics

Notes on these metrics:
* "Host" = M1 Macbook (arm64), "Docker" = Tutor MFE development container on said host
* For dev server times, the time of the second run of the command was measured; the first typically took twice as long while caches were filled.
* Maximum memory usage is the "maximum resident set size" measured with `/usr/bin/time -l`. The "peak memory footprint" reported by that command was even more favorable but I'm not sure which metric is better to use.
* The dependencies have only increased, not decreased, because frontend-platform still requires frontend-build as a peer dependency, so webpack and all of its dependencies still get installed.
* What's more, `frontend-component-header` depends on a bunch of legacy stuff that shouldn't be in the build - `babel-polyfill`, `babel-runtime`, and `core-js`. Fixing that will improve these metrics even further I think.

## Screenshots

The only difference in the result I could notice is a change to the default avatar because of how Vite currently handles SVG imports. There is probably a way to make it do the same thing that webpack was doing, but if you look at the code here you'll see that this way is actually simpler and more consistent than what was happening before.

| frontend-build | Vite |
|-|-|
|  frontend-build |  vite |


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react-jsx",
+
+    /* Linting */
+    "strict": true,
+    "noImplicitAny": false,  // Work around temporary lack of type data
```
bradenmacdonald, line 19: ^ Other than this `noImplicitAny` line, this file is unchanged from the Vite `react-ts` template.

tsconfig.node.json
```diff
@@ -0,0 +1,11 @@
+{
```
bradenmacdonald, line 1: This file is unchanged from the Vite `react-ts` template.

.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
bradenmacdonald, line 1: This file is unchanged from the Vite `react-ts` template.

mphilbrick211:
Hi @openedx/2u-aperture! Would someone mind taking a look at this? Thanks!

bradenmacdonald:
Note this is being discussed at https://discuss.openedx.org/t/experiment-build-an-mfe-with-vite-instead-of-frontend-build-webpack/12702

package.json
```diff
@@ -15,6 +15,7 @@
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "snapshot": "fedx-scripts jest --updateSnapshot",
     "start": "vite --host || echo this echo is here to ignore the --config arg if tutor passes it in",
+    "start-tutor": "PUBLIC_PATH=/profile/ MFE_CONFIG_API_URL='http://localhost:8000/api/mfe_config/v1' vite --host",
```
arbrandes, line 18: I'm _very_ interested in getting something like this working _before_ (and if) we migrate to Vite, haha.  I'll give it a shot and let you know.

bradenmacdonald: It works! I use it in course-authoring all the time. It's much easier and faster than running it inside a tutor container.

The command is:
```
APP_ID=course-authoring PUBLIC_PATH=/course-authoring/ MFE_CONFIG_API_URL='http://localhost:8000/api/mfe_config/v1' npm run start -- --host apps.local.overhang.io
```

And honestly the MFEs should be fixed so these env vars are the defaults, which would simplify this a lot.

deborahgu:
@bradenmacdonald, this is a proof of concept PR, right? Should it be closed?

bradenmacdonald:
@deborahgu Yes, that makes sense. It's not intended to be merged for now.

openedx-webhooks:
@bradenmacdonald Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-app-profile&entry.867055334=https://github.com/openedx/frontend-app-profile/pull/1000&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2024-04-04+19:35&entry.1917517419=2024-07-24+16:46&entry.2133058324=No) so we can improve your experience in the future.

PR #1017, opened 2024-05-10
feat: use frontend-plugin-framework to provide a FooterSlot

brian-smith-tcril:


jsnwesson (changes requested):
Just putting this request since I know this PR will eventually switch to using the `frontend-slot-footer` library and I don't want this to accidentally get merged.

package.json
```diff
@@ -38,6 +37,7 @@
     "@fortawesome/free-regular-svg-icons": "6.5.2",
     "@fortawesome/free-solid-svg-icons": "6.5.2",
     "@fortawesome/react-fontawesome": "0.2.0",
+    "@openedx/frontend-slot-footer": "^1.0.2",
```
arbrandes, line 40: Shouldn't we add an explicit FPF dependency here, as in the other MFEs?

arbrandes (approved):
Good to merge once comment is addressed.

package.json
brian-smith-tcril: It's a straight dependency in `frontend-slot-footer` https://github.com/openedx/frontend-slot-footer/blob/53987a478f615ec775137cc1bdfba982b8d0b3b6/package.json#L35 so I'm fine either way. If you'd prefer we have FPF explicitly set as a dependency in all the MFEs using `frontend-slot-footer` I'm more than happy to make that change.

arbrandes: Yeah, I guess adding a dependency if it's not used directly in the code is probably more confusing than otherwise.  Let's leave it out.

jsnwesson: agreed on leaving it out until we specifically need it for Profile and it reduces the number of dependencies to update for profile

PR #1033, opened 2024-06-13
build: Update codecov and use token

arbrandes:
Update codecov to the latest version and start using the org-wide token for uploads.

See https://github.com/openedx/wg-frontend/issues/179

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
     - run: make requirements
     - run: make test NPM_TESTS=build
     - run: make test NPM_TESTS=${{ matrix.npm-test }}
-    - name: upload coverage
-      uses: codecov/codecov-action@v3
+    - name: Coverage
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,7 +23,9 @@ jobs:
     - run: make requirements
     - run: make test NPM_TESTS=build
     - run: make test NPM_TESTS=${{ matrix.npm-test }}
-    - name: upload coverage
-      uses: codecov/codecov-action@v3
+    - name: Coverage
+      if: matrix.npm-test == 'test'
```
brian-smith-tcril, line 27:
```suggestion
      if: ${{ matrix.npm-test == 'test' }}
```

I think this is the syntax we need here (just going off of https://docs.github.com/en/actions/learn-github-actions/expressions

arbrandes: It seems our wiki was wrong, then.  Good catch.

justinhynes (approved):

